PR #11, opened 2023-08-04
Removed use of variables after move

ibhati:
Coverity caught some benign issues with using objects after moving. Replaced the usages with local/private objects.


hildebrandmw:
Thanks Ishwar!

hildebrandmw (approved):

marianotepper (approved):

PR #31, opened 2024-03-07
Allow cibuildwheel to be canceled by new commits.

hildebrandmw:
The recent [sync pr](#30) resulted in a bunch of commits with CI running on each commit. This is less than ideal and just wastes CI resources.

Most of our actions were set up to auto-cancel on new commit, but cibuildwheel was not.


ibhati (approved):

hildebrandmw:
Directly merging this PR as it only touches files that are not part of the import/export flow.

PR #46, opened 2024-07-18
Fix manylinux2014_x86_64 repo updating

mihaic:
See upstream issue:
https://github.com/pypa/manylinux/issues/1641


mihaic:
I will make the change with Copybara.

PR #47, opened 2024-07-18
Add dynamic Vamana index docs to Sphinx (#369)

mihaic:
Add dynamic Vamana index docs to Sphinx (#369)

Fixes #368


marianotepper (approved):
Everything looks good

PR #62, opened 2024-12-12
ci: add license header check

homksei:


.github/workflows/skywalking-eyes.yml
```diff
@@ -0,0 +1,41 @@
...
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref_name }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  check-license-header:
+    name: Copyright Check
+    runs-on: ubuntu-24.04
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
mihaic, line 37: How about using the tag as a refernce?
```suggestion
        uses: actions/checkout@v4.2.2
```

.github/workflows/skywalking-eyes.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: "Run check"
+        uses: apache/skywalking-eyes/header@cd7b195c51fd3d6ad52afceb760719ddc6b3ee91 # v0.6.0
```
mihaic, line 39:
```suggestion
        uses: apache/skywalking-eyes/header@v0.6.0
```

mihaic (changes requested):

.github/workflows/skywalking-eyes.yml
homksei: Using a hash rather than a version is a more desirable and secure [option](https://github.com/ossf/scorecard/blob/ea7e27ed41b76ab879c862fa0ca4cc9c61764ee4/docs/checks.md#pinned-dependencies)

.github/workflows/skywalking-eyes.yml
homksei: same as here https://github.com/intel/ScalableVectorSearch/pull/62#discussion_r1882892291

mihaic (approved):

.github/workflows/skywalking-eyes.yml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2023 Intel Corporation
```
napetrov, line 1: Should it be 2024?

homksei: yes, thanks.

homksei:
@Mergifyio rebase

PR #69, opened 2024-12-18
Expose Vamana build parameters for static and dynamic indices

ibhati:


bindings/python/src/vamana.cpp
```diff
@@ -486,7 +486,8 @@ void wrap(py::module& m) {
         .def_readwrite(
             "max_candidate_pool_size",
             &svs::index::vamana::VamanaBuildParameters::max_candidate_pool_size
-        );
+        )
+        .def_readwrite("prune_to", &svs::index::vamana::VamanaBuildParameters::prune_to);
```
mihaic, line 490: Are all the other changes in the C++ API covered by the Python API?

mihaic (changes requested):

bindings/python/src/vamana.cpp
ibhati: I left out "full_search_history" thinking that VamanaBuildParameters init with "true" value. But getting/setting this param is not a bad idea. I will add it in the Python API as well. thanks!

aguerreb (approved):

dian-lun-lin (approved):

mihaic (approved):

PR #71, opened 2025-01-17
Add timeout in batch iterator

ibhati:


mihaic (approved):

dian-lun-lin (approved):

rfsaliev:
According to Redis API, it wasn't necessary to add `cancel` argument to all iterator's calls - it was enough just get `cancel` functor at .ctor call, store in internal field and reuse it later.

PR #73, opened 2025-01-22
Add a new thread pool that switches between single- and multi- threaded execution

dian-lun-lin:
A thread pool that dynamically switches between single-threaded and multi-threaded execution.
- If `n == 1`, the task will be executed on the main thread without any locking mechanism that exists in `NativeThreadPool`.
- For `n > 1`, the tasks will be delegated to the internal `NativeThreadPool` for parallel execution.

### DIFF
--- a/include/svs/lib/threads/threadpool.h
+++ b/include/svs/lib/threads/threadpool.h
@@ -256,6 +256,31 @@ auto create_on_nodes(InterNUMAThreadPool& threadpool, F&& f)
 #endif
 
 /////
+///// A thread pool that dynamically switches between single-threaded and multi-threaded execution.
+///// - If `n == 1`, the task will be executed on the main thread without any locking mechanism.
+///// - For `n > 1`, the tasks will be delegated to the internal `NativeThreadPool` for parallel execution.
+/////
+class SwitchNativeThreadPool {
+  public:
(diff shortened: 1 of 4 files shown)


ibhati (approved):

include/svs/lib/threads/threadpool.h
```diff
@@ -255,6 +255,31 @@ auto create_on_nodes(InterNUMAThreadPool& threadpool, F&& f)
 }
 #endif
 
+/////
+///// A thread pool that dynamically switches between single-threaded and multi-threaded execution.
+///// - If `n == 1`, the task will be executed on the main thread without any locking mechanism.
+///// - For `n > 1`, the tasks will be delegated to the internal `NativeThreadPool` for parallel execution.
+/////
+class SwitchNativeThreadPool {
```
rfsaliev, line 263: According to the [`NativeThreadPoolBase ::parallel_for()`](https://github.com/intel/ScalableVectorSearch/blob/main/include/svs/lib/threads/threadpool.h#L189) implementation, a task is executed in the current thread in case of `n == 1`.
Why we cannot just fix existing code rather than making one more class?

rfsaliev: Ok, thanks. Looking to `NativeThreadPool::size()` code I did not realize that `NativeThreadPool` is thread-safe.

PR #75, opened 2025-01-29
add reuse_empty flag and update heuristics used in vamana construct

dian-lun-lin:
This PR includes two updates:

1. Introduces the `reuse_empty` flag, allowing users to decide whether to reuse empty entries that may exist after deletion and consolidation. Previously, SVS always scanned from the beginning of the memory to find empty entries, which could negatively impact performance. This change gives users the flexibility to optimize performance based on their specific needs. By default, the `reuse_empty` flag is set to false to optimize performance.
2. Updates the heuristics used in the Vamana construct. This addresses the issue where, even with fewer than 40 points, SVS would still iterate 40 times, causing unnecessary overhead. The new heuristics aim to reduce this inefficiency.

### DIFF
--- a/bindings/python/src/dynamic_vamana.cpp
+++ b/bindings/python/src/dynamic_vamana.cpp
@@ -92,7 +92,8 @@ template <typename ElementType>
 void add_points(
     svs::DynamicVamana& index,
     const py_contiguous_array_t<ElementType>& py_data,
-    const py_contiguous_array_t<size_t>& ids
+    const py_contiguous_array_t<size_t>& ids,
+    bool reuse_empty = false
 ) {
     if (py_data.ndim() != 2) {
         throw ANNEXCEPTION("Expected points to have 2 dimensions!");
@@ -105,7 +106,7 @@ void add_points(
             "Expected IDs to be the same length as the number of rows in points!"
         );
     }
-    index.add_points(data_view(py_data), std::span(ids.data(), ids.size()));
+    index.add_points(data_view(py_data), std::span(ids.data(), ids.size()), reuse_empty);
 }
 
 const char* ADD_POINTS_DOCSTRING = R"(
@@ -117,6 +118,8 @@ Args:
         the index.
     ids: Vector of ids to assign to each row in ``points``. Must have the same number of
         elements as ``points`` has rows.
+    reuse_empty: A flag that determines whether to reuse empty slots that may exist after deletion and consolidation.
+    If true, the scan starts from the beginning of the data to reuse empty slots.
 
 Furthermore, all entries in ``ids`` must be unique and not already exist in the index.
 If either of these does not hold, an exception will be thrown without mutating the
@@ -130,6 +133,7 @@ void add_points_specialization(py::class_<svs::DynamicVamana>& index) {
         &add_points<ElementType>,
         py::arg("points"),
         py::arg("ids"),
+        py::arg("reuse_empty") = false,
         ADD_POINTS_DOCSTRING
     );
 }
--- a/bindings/python/src/dynamic_vamana.cpp
+++ b/bindings/python/src/dynamic_vamana.cpp
@@ -118,12 +118,16 @@ Args:
         the index.
     ids: Vector of ids to assign to each row in ``points``. Must have the same number of
         elements as ``points`` has rows.
-    reuse_empty: A flag that determines whether to reuse empty slots that may exist after deletion and consolidation.
-    If true, the scan starts from the beginning of the data to reuse empty slots.
+    reuse_empty: A flag that determines whether to reuse empty entries that may exist after deletion and consolidation. When enabled,
+    scan from the beginning to find and fill these empty entries when adding new points.
 
 Furthermore, all entries in ``ids`` must be unique and not already exist in the index.
 If either of these does not hold, an exception will be thrown without mutating the
 underlying index.
+
+When ``delete_entries`` is called, a soft deletion is performed, marking the entries as ``deleted``.
+The state of these deleted entries becomes ``empty`` as the user calls ``consolidate``.
+When ``add_points`` is called with the ``reuse_empty`` flag enabled, the memory is scanned from the beginning to locate and fill these empty entries with new points.
 )";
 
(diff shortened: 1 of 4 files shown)


bindings/python/src/dynamic_vamana.cpp
```diff
@@ -92,7 +92,8 @@ template <typename ElementType>
 void add_points(
     svs::DynamicVamana& index,
     const py_contiguous_array_t<ElementType>& py_data,
-    const py_contiguous_array_t<size_t>& ids
+    const py_contiguous_array_t<size_t>& ids,
+    bool reuse_empty = false
```
mihaic, line 96: Users may be familiar with the similar argument in HNSWlib called `replace_deleted`. Should we use the same terminology?

ibhati (approved):

bindings/python/src/dynamic_vamana.cpp
```diff
@@ -117,6 +118,8 @@ each point.
         the index.
     ids: Vector of ids to assign to each row in ``points``. Must have the same number of
         elements as ``points`` has rows.
+    reuse_empty: A flag that determines whether to reuse empty slots that may exist after deletion and consolidation.
```
mihaic, line 121: I could only find the term `slot` once in the SVS documentation, in [`delete_entries`](https://intel.github.io/ScalableVectorSearch/cpp/internal/index_impl/dynamic_vamana.html#_CPPv4I0EN3svs5index6vamana18MutableVamanaIndex14delete_entriesEvRK1T), without a clear definition. I think it would be better to describe the new argument in general terms, without referring to slots. Furthermore, the [`consolidate`](https://intel.github.io/ScalableVectorSearch/python/dynamic.html#svs.DynamicVamana.consolidate) documentation mentions memory consumption, which seems to me related to the new argument. Can you please explain the interaction between `consolidate` and the new argument in a reply, not necessarily in the docstring?

mihaic (changes requested):

bindings/python/src/dynamic_vamana.cpp
dian-lun-lin: I'm a bit hesitant to use that term, as we don't actually replace deleted points unless we call consolidate. I'm concerned that if we use the term `replace_deleted`, users might assume that after calling `delete_entries`, the `add_points` will automatically replace those deleted entries if the flag is enabled (which is the case in HNSWlib). I use the term `empty` because, after consolidation, the states of those deleted entries become empty.

Do you think we should still use `replace_deleted`, or do you have other suggestions? Thanks.

bindings/python/src/dynamic_vamana.cpp
dian-lun-lin: When we call `delete_entries`, we do the soft deletion and mark those entries as `deleted`.
After consolidation, the state of those deleted entries will become empty. When we call `add_points` with the `reuse_empty` flag enabled, we will scan the memory from the beginning to find and fill these empty entries (i.e., slots).

Thanks for the suggestion. I will revise the documentation accordingly.

bindings/python/src/dynamic_vamana.cpp
```diff
@@ -117,10 +118,16 @@ each point.
...
         elements as ``points`` has rows.
+    reuse_empty: A flag that determines whether to reuse empty entries that may exist after deletion and consolidation. When enabled,
+    scan from the beginning to find and fill these empty entries when adding new points.
 
 Furthermore, all entries in ``ids`` must be unique and not already exist in the index.
 If either of these does not hold, an exception will be thrown without mutating the
 underlying index.
+
+When ``delete_entries`` is called, a soft deletion is performed, marking the entries as ``deleted``.
+The state of these deleted entries becomes ``empty`` as the user calls ``consolidate``.
```
mihaic, line 129:
```suggestion
When ``consolidate`` is called, the state of these deleted entries becomes ``empty``.
```

include/svs/index/vamana/dynamic_index.h
```diff
@@ -574,13 +578,24 @@ class MutableVamanaIndex {
 
     ///
     /// @brief Add the points with the given external IDs to the dataset.
+    //
+    /// When `delete_entries` is called, a soft deletion is performed, marking the entries
+    /// as `deleted`. The state of these deleted entries becomes `empty` as the user call
+    /// `consolidate`. When `add_points` is called with the `reuse_empty` flag enabled, the
```
mihaic, line 584:
```suggestion
    /// as `deleted`. When `consolidate` is called, the state of these deleted entries becomes `empty`.
    /// When `add_points` is called with the `reuse_empty` flag enabled, the
```

include/svs/index/vamana/vamana_build.h
```diff
@@ -219,6 +219,13 @@ class VamanaBuilder {
         size_t num_batches = std::max(
             size_t{40}, lib::div_round_up(num_nodes, lib::narrow_cast<size_t>(64 * 64))
         );
+
+        // If num_batches is greater than num_nodes (num_batches will be 40), set
```
mihaic, line 223: The reason is clear without the specific number. Plus, if we ever change `40` in the code, we will likely forget to change it in this comment. :)
```suggestion
        // If num_batches is greater than num_nodes, set
```

bindings/python/src/dynamic_vamana.cpp
mihaic: That makes sense. It is indeed different from HNSWlib, so it should not have the same name.

mihaic (changes requested):
Looks great! I left a couple of minor comments.

include/svs/index/vamana/vamana_build.h
mihaic: That makes sense. It is indeed different from HNSWlib, so it should not have the same name.

bindings/python/src/dynamic_vamana.cpp
mihaic: That makes sense. It is indeed different from HNSWlib, so it should not have the same name.

mihaic (approved):

PR #84, opened 2025-02-13
Allocator handle for custom allocators used in shared library

dian-lun-lin:
This PR is specifically designed to support a custom allocator that works with a shared library. It offers an interface that enables the implementation of a custom allocator not included in the shared library. Which custom allocator to use is determined at runtime.

This PR provides three utility functions:
1. `make_allocator_handle(Alloc alloc)`: Accepts an allocator and returns an `AllocatorHandle` to wrap the allocator.
2. `make_blocked_allocator_handle(Alloc alloc)`: Accepts an allocator, wraps it in an `AllocatorHandle`, and returns a `Blocked` object that manages the allocator.
3. `make_blocked_allocator_handle(const BlockingParameters& parameters, Alloc alloc)`: Accepts an allocator, wraps it in an `AllocatorHandle`, and returns a `Blocked` object configured with the specified parameters.

The `AllocatorHandle` provides `allocate` and `deallocate` functions to use the wrapped custom allocator for memory allocation.

### DIFF
--- a/include/svs/core/allocator.h
+++ b/include/svs/core/allocator.h
@@ -515,7 +515,8 @@ concept HasValueType = requires { typename Alloc::value_type; };
 
 // clang-format off
 template <typename Alloc>
-concept Allocator = HasValueType<Alloc> && requires(Alloc& alloc, typename Alloc::value_type* ptr, size_t n) {
+concept Allocator = HasValueType<Alloc> && std::is_copy_constructible_v<Alloc>
+&& requires(Alloc& alloc, typename Alloc::value_type* ptr, size_t n) {
     { alloc.allocate(n) } -> std::same_as<typename Alloc::value_type*>;
 
     { alloc.deallocate(ptr, n) };
@@ -564,15 +565,16 @@ template <typename T> class AllocatorHandle {
     template <detail::Allocator Impl>
     explicit AllocatorHandle(Impl&& impl)
         requires(!std::is_same_v<Impl, AllocatorHandle>) &&
-                std::is_rvalue_reference_v<Impl&&>
-        : impl_{std::make_unique<AllocatorImpl<Impl>>(std::move(impl))} {}
+                std::is_rvalue_reference_v<Impl&&> &&
+                std::is_same_v<typename Impl::value_type, T>
+        : impl_{new AllocatorImpl(std::move(impl))} {}
 
     AllocatorHandle() {}
     AllocatorHandle(const AllocatorHandle& other)
         : impl_{other.impl_->clone()} {}
     AllocatorHandle(AllocatorHandle&&) = default;
     AllocatorHandle& operator=(const AllocatorHandle& other) {
-        impl_ = other.impl_->clone();
+        impl_.reset(other.impl_->clone());
         return *this;
     }
     AllocatorHandle& operator=(AllocatorHandle&&) = default;
--- a/include/svs/core/allocator.h
+++ b/include/svs/core/allocator.h
@@ -601,8 +601,8 @@ template <typename T> class AllocatorHandle {
 ///
 /// AllocatorHandle
 /// ================
-/// `AllocatorHandle` is specifically designed to support LVQ/Leanvec Datasets with a custom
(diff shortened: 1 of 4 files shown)


include/svs/core/allocator.h
```diff
@@ -507,4 +507,121 @@ class MemoryMapper {
...
+  private:
+    std::unique_ptr<AllocatorInterface> impl_;
+};
+
+///
+/// @class allocator_handle
+///
+/// AllocatorHandle
+/// ================
+/// `AllocatorHandle` is specifically designed to support LVQ/Leanvec Datasets with a custom
```
ibhati, line 604: Is there something unique about the AllocatorHandle when it comes to LVQ/Leanvec datasets? From what I understand, its primary use case is to enable these custom datasets to work with a shared library. However, I believe that the AllocatorHandle is more versatile and can be applied to a variety of other scenarios as well. Is that right @dian-lun-lin ?

include/svs/core/data/simple.h
```diff
@@ -826,4 +826,44 @@ template <typename T, size_t Extent = Dynamic, typename Alloc = lib::Allocator<T
...
+    return data::Blocked{make_allocator_handle(std::move(alloc))};
+}
+
+///
+/// @brief Creates a `Blocked` object with an `AllocatorHandle` from a given allocator.
+///
+/// @tparam Alloc The type of the allocator, which must satisfy the `Allocator` concept.
+/// @param alloc The allocator to be wrapped in an `AllocatorHandle and managed by a
+/// `Blocked` object.
+//  @param parameters The blocking parameters used to configure the `Blocked` object.
```
ibhati, line 855: I think the order of the arguments in the doc string is reversed. First parameters and then alloc?

tests/svs/core/allocator.cpp
```diff
@@ -137,4 +137,64 @@ CATCH_TEST_CASE("Testing Allocator", "[allocators]") {
             }
         }
     }
+
+    CATCH_SECTION("Testing `AllocatorHandle`") {
```
ibhati, line 141: Should we add test for Blocked AllocatorHandle as well?

include/svs/core/allocator.h
dian-lun-lin: Yes, `AllocatorHandle` can be applied to a broader range of scenarios. I initially focused on its use with LVQ/LeanVec, but I will update the documentation to reflect its applicability. Thank you!

ibhati (approved):

include/svs/core/allocator.h
```diff
@@ -507,4 +507,121 @@ class MemoryMapper {
...
+        return *this;
+    }
+    AllocatorHandle& operator=(AllocatorHandle&&) = default;
+    ~AllocatorHandle() = default;
+
+    T* allocate(size_t n) {
+        if (impl_.get() == nullptr) {
+            throw ANNEXCEPTION("Empty allocator handle!");
+        }
+        return static_cast<T*>(impl_->allocate(n));
```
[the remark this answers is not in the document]
dian-lun-lin: That's a good suggestion. I've added the requirement in the constructor. Thank you.

include/svs/core/allocator.h
```diff
@@ -507,4 +507,121 @@ class MemoryMapper {
...
+        requires(!std::is_same_v<Impl, AllocatorHandle>) &&
+                std::is_rvalue_reference_v<Impl&&>
+        : impl_{std::make_unique<AllocatorImpl<Impl>>(std::move(impl))} {}
+
+    AllocatorHandle() {}
+    AllocatorHandle(const AllocatorHandle& other)
+        : impl_{other.impl_->clone()} {}
+    AllocatorHandle(AllocatorHandle&&) = default;
+    AllocatorHandle& operator=(const AllocatorHandle& other) {
+        impl_ = other.impl_->clone();
```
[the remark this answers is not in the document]
dian-lun-lin: I've updated the code to use `reset()`. Thanks for catching that!

PR #88, opened 2025-02-24
fix: add alias for IntelLLVM compiler

ahuber21:
This adds the correct `archspec` name for `CMake` `IntelLLVM`.

### DIFF
--- a/bindings/python/include/svs/python/manager.h
+++ b/bindings/python/include/svs/python/manager.h
@@ -119,7 +119,9 @@ reconstruct(Index& index, py_contiguous_array_t<uint64_t> ids) {
     auto destination = py_contiguous_array_t<float>({num_ids, data_dims});
     index.reconstruct_at(
         mutable_data_view(destination),
-        std::span<const uint64_t>(ids.template mutable_unchecked().mutable_data(), num_ids)
+        std::span<const uint64_t>(
+            ids.template mutable_unchecked<-1>().mutable_data(), num_ids
+        )
     );
 
     // Reshape the destination to have the same shape as the original IDs (plus the extra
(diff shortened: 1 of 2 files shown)


bindings/python/microarch.py
```diff
@@ -85,6 +85,7 @@ def resolve_compiler(name: str):
     aliases = {
         "GNU": "gcc",
         "Clang": "clang",
+        "IntelLLVM": "intel"
```
mihaic, line 88: I think `intel` is the Intel® C++ Compiler Classic and that the name should be `oneapi` unless I misunderstood your setup.
https://github.com/archspec/archspec-json/pull/47
```suggestion
        "IntelLLVM": "oneapi"
```

mihaic (changes requested):

bindings/python/microarch.py
ahuber21: Yes you are right.

bindings/python/include/svs/python/manager.h
```diff
@@ -119,7 +119,9 @@ reconstruct(Index& index, py_contiguous_array_t<uint64_t> ids) {
     auto destination = py_contiguous_array_t<float>({num_ids, data_dims});
     index.reconstruct_at(
         mutable_data_view(destination),
-        std::span<const uint64_t>(ids.template mutable_unchecked().mutable_data(), num_ids)
+        std::span<const uint64_t>(
+            ids.template mutable_unchecked<-1>().mutable_data(), num_ids
```
ahuber21, line 123: For using `oneapi` compiler it is necessary to provide the `<-1>` (which is the default value for the `Dims` template parameter in `mutable_unchecked()` to avoid a compiler warning that gets promoted to an error

```
manager.h:122:48: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
```

mihaic (approved):

PR #97, opened 2025-03-21
feat: Add scalar quantization dataset

ahuber21:
This PR adds a new class `SQDataset` which implements global scalar quantization.


examples/cpp/vamana.cpp
```diff
@@ -156,6 +158,36 @@ int svs_main(std::vector<std::string> args) {
     index.set_threadpool(svs::threads::DefaultThreadPool(4));
     //! [Set a new thread pool with n-threads]
 
+    //! [Compressed Loader]
+    // Quantization
+    namespace scalar = svs::quantization::scalar;
+
+    // Wrap the compressor object in a lazy functor.
+    // This will defer loading and compression of the LVQ dataset until the threadpool
```
ibhati, line 166:
```suggestion
    // This will defer loading and compression of the SQ dataset until the threadpool
```

dian-lun-lin:
For dynamic vamana, we need to update global maximum and minimum when we add/delete points. Currently, this implementation only finds global maximum and minimum when calling `compress`.

ahuber21:
@dian-lun-lin added

* A check for trivial data which doesn't really make sense with compression: https://github.com/intel/ScalableVectorSearch/blob/d18182c33ff7e7426d9542e74485096d7ca2c5e9/tests/svs/quantization/scalar/scalar.cpp#L324-L337
* A check for re-calculation of scale and bias if necessary https://github.com/intel/ScalableVectorSearch/blob/d18182c33ff7e7426d9542e74485096d7ca2c5e9/tests/svs/quantization/scalar/scalar.cpp#L344-L358
* A naive implementation of adaptive recompression. We only need to recompress if the global min/max values in the dataset change. https://github.com/intel/ScalableVectorSearch/blob/d18182c33ff7e7426d9542e74485096d7ca2c5e9/include/svs/quantization/scalar/scalar.h#L471-L488

dian-lun-lin:
> @dian-lun-lin added
> 
> * A check for trivial data which doesn't really make sense with compression: https://github.com/intel/ScalableVectorSearch/blob/d18182c33ff7e7426d9542e74485096d7ca2c5e9/tests/svs/quantization/scalar/scalar.cpp#L324-L337
> * A check for re-calculation of scale and bias if necessary https://github.com/intel/ScalableVectorSearch/blob/d18182c33ff7e7426d9542e74485096d7ca2c5e9/tests/svs/quantization/scalar/scalar.cpp#L344-L358
> * A naive implementation of adaptive recompression. We only need to recompress if the global min/max values in the dataset change. https://github.com/intel/ScalableVectorSearch/blob/d18182c33ff7e7426d9542e74485096d7ca2c5e9/include/svs/quantization/scalar/scalar.h#L471-L488

Should we allow the insertion of trivial data? Upstream applications might rely on it for unit testing. Blocking such data could disrupt their tests.

ahuber21:
Thank you @razdoburdin for the PR against this branch. I've checked my usage of `eve` and I don't think I should have used it in the `fix_argument()` function. I've replaced it with a simple `std::reduce()` and thereby removed the necessity to `#include eve/algo.h`, which hopefully fixes the build error on ARM.

@ibhati @dian-lun-lin as discussed I reverted the re-compression changes (and the check for trivial data). My last action for this PR would be clean up the RNG generator in `scalar.cpp`, which I'm hoping to finish before your AM so that we can merge quickly.

include/svs/quantization/scalar/scalar.h
```diff
@@ -0,0 +1,475 @@
...
+
+        // Compress data
+        auto compressor = detail::Compressor<element_type, data_type>{scale, bias};
+        auto compressed = compressor(data, threadpool, allocator);
+
+        return SQDataset<element_type, extent, allocator_type>{
+            std::move(compressed), scale, bias};
+    }
+
+    /// @brief Compact the dataset
```
ibhati, line 431: Just to keep in mind we need to dd the Doxygen docstrings for the other functions and some more documentation for SQDataset in next PR

tests/integration/vamana/scalar_build.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+template <typename T, typename Distance> void test_build(const Distance& distance) {
+    // How far these results may deviate from previously generated results.
+    const double epsilon = 0.01;
+    const auto queries = svs::data::SimpleData<float>::load(test_dataset::query_file());
+    CATCH_REQUIRE(svs_test::prepare_temp_directory());
+    size_t num_threads = 2;
+
+    // use uncompressed reference results which should be identical
```
ibhati, line 78:
```suggestion
    // Use uncompressed reference results which should be identical
```

tests/svs/quantization/scalar/scalar.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    // Assert compressed data has the same size and dimensions
+    CATCH_REQUIRE(sq_dataset.size() == original.size());
+    CATCH_REQUIRE(sq_dataset.dimensions() == original.dimensions());
+
+    // Assert scale and bias are calculated correctly
+    // Scale is calculated from (max_data - min_data) / (max_quant - min_quant)
+    // The dataset features values [-127, 127], the quantization range is given by the MIN
+    // and MAX elements of the provided type.
+    constexpr float MIN = std::numeric_limits<T>::min();
```
ibhati, line 44: Is not the minimium value for int8 -128 instead of -127?

tests/svs/quantization/scalar/scalar.cpp
```diff
@@ -0,0 +1,211 @@
...
+    constexpr float MAX = std::numeric_limits<T>::max();
+    constexpr float exp_scale = 254.0F / float(MAX - MIN);
```
ibhati, line 46:
```suggestion
    constexpr float exp_scale = 255.0F / float(MAX - MIN);
```

tests/svs/quantization/scalar/scalar.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // Bias is calculated as min_data - min_quant * scale
+    constexpr float exp_bias = -127.0F - MIN * exp_scale;
```
ibhati, line 48:
```suggestion
    constexpr float exp_bias = -128.0F - MIN * exp_scale;
```

tests/svs/quantization/scalar/scalar.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // Calculations are performed in float everywhere and should therefore produce the exact
+    // same results
+    CATCH_REQUIRE(sq_dataset.get_scale() == exp_scale);
+    CATCH_REQUIRE(sq_dataset.get_bias() == exp_bias);
```
ibhati, line 52: Why are these checks not failing if the int8 minimum is -128? Not sure

tests/svs/quantization/scalar/scalar.cpp
ahuber21: Yes, but the dataset is symmetric, not sure why, but it doesn't cause issues.

tests/svs/quantization/scalar/scalar.cpp
ahuber21: No the dataset min max max values are -127 and 127, respectively. I could have used `min()` and `max()`, but figured I would use this knowledge about our test data in the test case.

tests/svs/quantization/scalar/scalar.cpp
ahuber21: Yeah as explained, these are the values that are not determined from the compression range but rather from data. So all good!

include/svs/quantization/scalar/scalar.h
```diff
@@ -0,0 +1,475 @@
...
+    MinMaxAccumulator operator()(const Dataset& data, Pool& threadpool) {
+        static constexpr size_t batch_size = 512;
+
+        // Thread-local accumulators
+        std::vector<MinMaxAccumulator> tls(threadpool.size());
+
+        // Compute mean and squared sum
+        threads::parallel_for(
+            threadpool,
+            threads::DynamicPartition(data.size(), batch_size),
```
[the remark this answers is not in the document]
ibhati: Can we use StaticPartion here as well?

ibhati (commented):
Looks good, just see if we can use StaticParition for the other instance as well

ahuber21:
@ibhati everything should be there now. Please update your "change requested" if you think we're ready to merge.

ibhati (approved):
Thanks @ahuber21 for enabling this capability in SVS.

PR #105, opened 2025-04-10
Refactor Vamana Iterator: simplified API and performance enhancement

ibhati:
This PR simplifies the Vamana Iterator API and enhances its performance by improving buffer management and search behavior.

*Key Changes:*

**Eliminated `soft_clear()` with Extra Buffer Capacity:**
The buffer size is now larger than the search_window_size, removing the need for `soft_clear()` when calling `next()`.
This ensures smoother search continuation and better performance.

**Search Triggered Exclusively by `next(batch_size)`:**
The iterator is now empty upon creation or after being updated.
Searches are only performed when `next(batch_size)` is explicitly called, providing better control over search execution. On calling `next(batch_size)`, both the `search_window_size` and `buffer_capacity` are increased by `batch_size` and then search is conducted.

**Dynamic Buffer Adjustment with batch_size:**
Introduced the `batch_size` argument in `next()`, allowing the search buffer size to be dynamically adjusted based on the batch size.

**Simplified API by Removing Schedules:**
Removed the usage of schedules in the iterator, simplifying its design and making it easier to use.


include/svs/index/vamana/iterator.h
```diff
@@ -244,15 +208,15 @@ class BatchIterator {
...
+    /// The span is invalidated by calls to `next(...)`.
     std::span<const value_type> contents() const { return lib::as_const_span(results_); }
 
-    /// @brief Return the number of buffered results.
+    /// @brief Returns the number of buffered results.
     size_t size() const { return results_.size(); }
 
-    /// @brief Return the current batch number contained in the buffer.
+    /// @brief Reurn the current batch nu contained in the buffermber contained in the
+    /// buffer.
```
mihaic, line 219:
```suggestion
    /// @brief Return the batch number corresponding to the current buffer.
```

mihaic, line 219: Have you considered a more explicit name for this method, such as `batch_number` or something similar? When I see just `batch` I think of the data in the buffer itself, not an iteration number.

mihaic (changes requested):
A few documentation fixes and a suggestion.

include/svs/orchestrators/vamana_iterator.h
```diff
@@ -159,14 +131,21 @@ class VamanaIterator {
         return impl_->results();
     }
 
-    /// @brief Retrieve a new batch of results.
+    /// @brief Prepate a new batch of results.
```
mihaic, line 134:
```suggestion
    /// @brief Prepare a new batch of results.
```

mihaic (changes requested):
Just a typo.

include/svs/orchestrators/vamana_iterator.h
ibhati: Oops!

include/svs/index/vamana/iterator.h
```diff
@@ -244,88 +211,53 @@ class BatchIterator {
...
     std::span<const value_type> contents() const { return lib::as_const_span(results_); }
 
-    /// @brief Return the number of buffered results.
+    /// @brief Returns the number of buffered results.
     size_t size() const { return results_.size(); }
 
-    /// @brief Return the current batch number contained in the buffer.
-    size_t batch() const { return iteration_; }
+    /// @brief Return the batch number corresponding to the current buffer.
+    /// buffer.
```
mihaic, line 222:
```suggestion
```

mihaic (changes requested):
One more typo.

examples/cpp/vamana_iterator.cpp
```diff
@@ -68,42 +68,34 @@ void check(bool value, std::string_view expr, svs::lib::detail::LineInfo linfo)
...
-    // After the constructor returns, the contents of the first batch will be available.
     auto itr = [&]() {
         // Construct a query a query in a scoped block to demonstrate that the iterator
         // maintains an internal copy.
         auto query = std::vector<float>{3.25, 3.25, 3.25, 3.25};
 
-        // Make a batch iterator for the query using the provided schedule.
-        return index.batch_iterator(svs::lib::as_const_span(query), schedule);
+        // Make a batch iterator for the query.
+        return svs::VamanaIterator(index, svs::lib::as_const_span(query));
```
mihaic, line 91: Since the example will not show `index.batch_iterator` anymore, we should mention it somewhere else in the docs (aside from the method doc) because it is still useful and users should be aware of it. This could be a separate PR.

mihaic (changes requested):

include/svs/index/vamana/iterator.h
```diff
@@ -47,181 +50,145 @@ template <std::integral I> struct RestartInitializer {
...
-    /// A random-access, contiguous iterator to ``cosnt value_type`` over the current batch
-    /// of results.
+    /// Random-access iterator to `const value_type` over the current batch of results.
     using const_iterator = typename result_buffer_type::const_iterator;
 
+    /// @brief Constructs a batch iterator for the given query over the index.
+    /// @param parent The index to search.
+    /// @param query The query data.
+    /// @param extra_search_buffer_capacity Additional buffer capacity for the search
+    ///     When not provided, default value SVS_ITERATOR_EXTRA_BUFFER_CAPACITY_DEFAULT is
```
mihaic, line 157: I think `SVS_ITERATOR_EXTRA_BUFFER_CAPACITY_DEFAULT` is used literally, not replaced by 100 by Doxygen. If we can make the definition of `SVS_ITERATOR_EXTRA_BUFFER_CAPACITY_DEFAULT` show up in the HTML docs, that's great. If not, we should just use `100` here.

In any case the same solution for presenting the default should be used here and in the `batch_iterator` methods.

mihaic (changes requested):

include/svs/index/vamana/iterator.h
ibhati: I gave it a shot, but pulling the values of these global variables/defines in C++ with Doxygen/Sphinx isn't as simple as generating Python docs. I'd rather stick with using the variable name instead of hardcoded values in docs, since hardcoded values can easily become outdated or inconsistent from the source code

include/svs/orchestrators/vamana_iterator.h
```diff
@@ -123,33 +98,30 @@ class VamanaIterator {
...
     /// @brief Construct a new batch iterator for the query over the index.
     ///
-    /// The results for the first batch will be available once the constructor completes.
-    ///
-    /// Argument ``schedule`` will be used to adjust the search parameters for each batch.
-    /// ``cancel`` is an optional argument to determine if the search should be cancelled.
-    /// An internal copy of the query will be made.
+    /// Argument ``extra_search_buffer_capacity`` is the extra search buffer capacity to use
+    /// for the next search. This is used to ensure that we have few extra neighbors in the
+    /// search buffer to accommodate the next search.
```
mihaic, line 103: The `extra_search_buffer_capacity` default should also be documented here.

include/svs/index/vamana/iterator.h
mihaic: I agree, it is better to not hard code the value in the docs. Nevertheless, we should show the value in the docs; users should not have to look at the source code to learn that value is 100. Looking at the HTML docs, I don't see 100.

ibhati: Yeah, created issue #118 to properly get the values of these globals

mihaic (approved):

PR #108, opened 2025-04-14
Fix python examples fail

### LINKED ISSUE #107: Python example failing
When running the Python example, I get the following assertion error:
```
~/ScalableVectorSearch$ python examples/python/example_vamana.py
Recall = 0.817
Traceback (most recent call last):
  File "/raid6/ishwarsi/projects/ScalableVectorSearch/examples/python/example_vamana.py", line 245, in <module>
    run()
  File "/raid6/ishwarsi/projects/ScalableVectorSearch/examples/python/example_vamana.py", line 162, in run
    assert(recall == 0.8288)
           ^^^^^^^^^^^^^^^^
AssertionError
```

This is probably introduced with #96 when the default parameters are added. We need to add some error margins.
Better to include running these examples in the CI with public repos.

yuejiaointel:
Use margin check and add tests to pipeline.

Fixes #107.


examples/python/example_vamana_dynamic.py
```diff
@@ -118,7 +121,9 @@ def run():
     # Compare with the groundtruth.
     recall = svs.k_recall_at(groundtruth, I, 10, 10)
     print(f"Recall = {recall}")
-    assert(recall == 0.8202)
+    expected_recall = 0.8202
+    assert recall < expected_recall + epsilon
+    assert recall > expected_recall - epsilon
```
mihaic, line 126: If we are keeping the `assert_equal` function, this looks like a good place to use it.
```suggestion
    assert_equal(recall, 0.8202)
```

examples/python/example_vamana_dynamic.py
```diff
@@ -158,7 +163,9 @@ def run():
...
```
mihaic, line 168:
```suggestion
    assert_equal(recall, 0.8202)
```

examples/python/example_vamana_dynamic.py
```diff
@@ -22,11 +22,12 @@
...
 
 DEBUG_MODE = False
-def assert_equal(lhs, rhs, message: str = ""):
+def assert_equal(lhs, rhs, message: str = "", epsilon = 0.05):
     if DEBUG_MODE:
         print(f"{message}: {lhs} == {rhs}")
     else:
-        assert lhs == rhs, message
+        assert lhs < rhs + epsilon, f"{message}"
+        assert lhs > rhs - epsilon, f"{message}"
```
mihaic, line 30: Do we need the f-string?
```suggestion
        assert lhs < rhs + epsilon, message
        assert lhs > rhs - epsilon, message
```

examples/python/example_vamana.py
```diff
@@ -21,11 +21,12 @@
...
 
 DEBUG_MODE = False
-def assert_equal(lhs, rhs, message: str = ""):
+def assert_equal(lhs, rhs, message: str = "", epsilon = 0.05):
     if DEBUG_MODE:
         print(f"{message}: {lhs} == {rhs}")
     else:
-        assert lhs == rhs, message
+        assert lhs < rhs + epsilon, f"{message}"
+        assert lhs > rhs - epsilon, f"{message}"
```
mihaic, line 29: Please apply the same suggestions from the dynamic example.

mihaic (changes requested):

examples/python/example_vamana_dynamic.py
yuejiaointel: That's a great idea, fixed!

examples/python/example_vamana_dynamic.py
yuejiaointel: Fixed, thx!

examples/python/example_vamana_dynamic.py
yuejiaointel: Good catch, fixed!

examples/python/example_vamana.py
yuejiaointel: Fixed, thx!

mihaic (changes requested):

.github/workflows/build-python.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    # Make sure to add the location of the generated wheel to the python path.
+    - name: Run tests
+      env:
+          PYTHONPATH: ${{ runner.temp }}/usr
+          CTEST_OUTPUT_ON_FAILURE: 1
+      working-directory: ${{ runner.temp }}
+      run: python -m unittest discover -s ${GITHUB_WORKSPACE}/bindings/python
+
+    - name: Run examples
```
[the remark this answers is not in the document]
yuejiaointel: Sure, added in cibuildwheel instead Thx!

mihaic (approved):

PR #111, opened 2025-04-17
Add MacOS validation to CI

razdoburdin:
This PR adds MacOS validation with clang in github workflow. Some changes in the codebase and tests are also made for compatibility with MacOS.

### DIFF
--- a/.github/workflows/build-macos.yaml
+++ b/.github/workflows/build-macos.yaml
@@ -1,0 +1,85 @@
+# Copyright 2025 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: MacOS Build and Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
+# This allows a subsequently queued workflow run to interrupt previous runs
+concurrency:
+  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: ${{ matrix.cxx }}, ${{ matrix.build_type }}
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        build_type: [RelWithDebugInfo]
+        cxx: [clang++-15]
+        include:
+          - cxx: clang++-15
+            package: llvm@15
+            cc_name: clang
+            cxx_name: clang++
+            needs_prefix: true
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install Compiler
+      run: |
+        echo "Installing ${{ matrix.package }}..."
+        brew install ${{ matrix.package }}
+
+    - name: Configure build
+      working-directory: ${{ runner.temp }}
+      env:
+        TEMP_WORKSPACE: ${{ runner.temp }}
+      run: |
+        if [[ "${{ matrix.needs_prefix }}" == "true" ]]; then
+          # For non-default packages like llvm@15, get the install prefix
+          COMPILER_PREFIX=$(brew --prefix ${{ matrix.package }})
+          export CC="${COMPILER_PREFIX}/bin/${{ matrix.cc_name }}"
+          export CXX="${COMPILER_PREFIX}/bin/${{ matrix.cxx_name }}"
+        else
+          # For versioned GCC installs, the name is usually directly available
+          export CC="${{ matrix.cc_name }}"
+          export CXX="${{ matrix.cxx_name }}"
+        fi
+
+        cmake -B${TEMP_WORKSPACE}/build -S${GITHUB_WORKSPACE} \
+              -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+              -DSVS_BUILD_BINARIES=YES \
+              -DSVS_BUILD_TESTS=YES \
+              -DSVS_BUILD_EXAMPLES=YES \
+              -DSVS_EXPERIMENTAL_LEANVEC=YES
+
+    - name: Build Tests and Utilities
+      working-directory: ${{ runner.temp }}/build
+      run: make -j$(sysctl -n hw.ncpu)
+
+    - name: Run tests
+      env:
+          CTEST_OUTPUT_ON_FAILURE: 1
+      working-directory: ${{ runner.temp }}/build/tests
+      run: ctest -C ${{ matrix.build_type }}
--- a/include/svs/core/allocator.h
+++ b/include/svs/core/allocator.h
@@ -41,7 +41,11 @@
 #include <array>
 #include <fcntl.h>
 #include <filesystem>
+
+#if defined(__linux__)
 #include <linux/mman.h>
+#endif // __linux__
+
 #include <memory>
 #include <optional>
 #include <string>
@@ -78,11 +82,17 @@ struct HugepageX86Parameters {
 };
 
 // Hugepage Allocation will happen in the order given below.
+#if defined(__linux__)
 static constexpr std::array<HugepageX86Parameters, 3> hugepage_x86_options{
     HugepageX86Parameters{1 << 30, MAP_HUGETLB | MAP_HUGE_1GB},
     HugepageX86Parameters{1 << 21, MAP_HUGETLB | MAP_HUGE_2MB},
     HugepageX86Parameters{1 << 12, 0},
 };
+#else
+static constexpr std::array<HugepageX86Parameters, 1> hugepage_x86_options{
+    HugepageX86Parameters{1 << 12, 0}
+};
+#endif // __linux__
 
 namespace detail {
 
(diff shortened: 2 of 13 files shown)


include/svs/core/allocator.h
```diff
@@ -78,11 +82,17 @@ struct HugepageX86Parameters {
 };
 
 // Hugepage Allocation will happen in the order given below.
+#if defined(__linux__)
```
ahuber21, line 85: Please add a comment here that explains the differences

include/svs/core/allocator.h
```diff
@@ -78,11 +82,17 @@ struct HugepageX86Parameters {
...
 static constexpr std::array<HugepageX86Parameters, 3> hugepage_x86_options{
     HugepageX86Parameters{1 << 30, MAP_HUGETLB | MAP_HUGE_1GB},
     HugepageX86Parameters{1 << 21, MAP_HUGETLB | MAP_HUGE_2MB},
     HugepageX86Parameters{1 << 12, 0},
 };
+#else
+static constexpr std::array<HugepageX86Parameters, 1> hugepage_x86_options{
```
ahuber21, line 92: I am a bit confused that changes to `HugepageX86Parameters` are necessary for MacOS validation. Is the validation Intel MacOS too or only Apple Silicon?
In case it targets Intel MacOS too I'm wondering if we really need it. By now all these machines are pretty old and is it really worth adding the extra complexity for superseded hardware?
In case this is also necessary or Apple Silicon, could you explain why? Is there now a mixture between x86 and ARM64?

include/svs/lib/threads/thunks.h
```diff
@@ -104,7 +104,8 @@ template <typename F, typename I> struct Thunk<F, DynamicPartition<I>> {
                     return;
                 }
 
-                auto stop = std::min(grainsize * (i + 1), iterator_size);
+                auto stop = std::min(grainsize * (i + 1),
```
ahuber21, line 107: Technically this should be a separate PR.

tests/integration/inverted/build.cpp
```diff
@@ -64,7 +64,12 @@ void run_test(const Queries& queries, ThreadPoolProto threadpool_proto) {
     auto strategy = Strategy();
 
     // Distance between the obtained results and reference ressults.
-    const double epsilon = 0.005;
+    #if defined(__APPLE__)
```
ahuber21, line 67: I don't like accepting different accuracies for different architectures. Let's discuss this with @ibhati

tests/svs/lib/threads/threadpool.cpp
```diff
@@ -117,7 +117,7 @@ CATCH_TEST_CASE("Thread Pool", "[core][threads][threadpool]") {
                 [&](const auto& range, uint64_t tid) {
                     std::lock_guard lock{mutex};
                     seen_threads.push_back(tid);
-                    ranges.push_back(threads::UnitRange<size_t>{
+                    ranges.push_back(threads::UnitRange<uint64_t>{
```
ahuber21, line 120: Please comment why this is necessary.

tests/svs/lib/timing.cpp
```diff
@@ -45,6 +45,15 @@ void stress(svs::lib::Timer& timer, size_t max_depth = 3, size_t max_flat = 3) {
         }
     }
 }
+
+// std::this_thread::sleep_for doesn't provide good accuracy for macos
```
ahuber21, line 49: This comment can be improved. If your changes get merged all traces of `std::this_thread::sleep_for` are gone and it's confusing. Rather explain the rationale of using a busy sleep.
Would also like to hear @ibhati's opinion.

ahuber21 (changes requested):
Thanks! Some minor comments. With enabling more architectures we really must be careful that things don't explode in complexity.
I would like to keep the architecture and OS specific complexity out of the existing header files as much as possible.
So whenever possible put them into a separate file. It could be worth exploring if adding a compatibility directory that provides headers for exactly that makes things more understandable.

tests/svs/core/allocator.cpp
```diff
@@ -42,7 +42,11 @@ static_assert(std::is_same_v<typename Traits::pointer, float*>);
 static_assert(std::is_same_v<typename Traits::const_pointer, const float*>);
 static_assert(std::is_same_v<typename Traits::void_pointer, void*>);
 static_assert(std::is_same_v<typename Traits::const_void_pointer, const void*>);
+#if defined(__APPLE__)
```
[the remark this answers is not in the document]
razdoburdin: There are some differences in integer types for MacOS (I guess it is a legacy).

tests/svs/index/inverted/clustering.cpp
```diff
@@ -412,8 +412,8 @@ CATCH_TEST_CASE("Clustering with Logger", "[logging]") {
     auto clustering_parameters = svs::index::inverted::ClusteringParameters()
                                      .percent_centroids(svs::lib::Percent(0.1))
                                      .epsilon(0.05)
-                                     .max_replicas(8)
-                                     .max_cluster_size(200);
+                                     .max_replicas(12)
```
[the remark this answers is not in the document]
razdoburdin: The original test failed on MacOS. I suspect, the reason is platform-specific differences in random number generation, so I added a bit more flexibility to pruning.

tests/svs/lib/threads/threadpool.cpp
razdoburdin: For MacOS `size_t` is not the same as `uint64_t`. We can't not push the `size_t` to the vector of `uint64_t`.

include/svs/lib/threads/thunks.h
razdoburdin: It is the part of codebase adaptation for MacOS. The `size_t` for Mac is technically not the same type as `uint64_t` even for 64 bits. So we need to manually cast `iterator_size` from `size_t` to `uint64_t` to help compiler to deduce template parameter for `std::min`.

include/svs/core/allocator.h
razdoburdin: The optimized huge-page processing is available for linux only (doesn't matter Intel MacOS or Apple Silicon), so we need to isolate this OS-specific code. I added some comments in the code, to make it more clear.

razdoburdin:
> With enabling more architectures we really must be careful that things don't explode in complexity. I would like to keep the architecture and OS specific complexity out of the existing header files as much as possible. So whenever possible put them into a separate file. It could be worth exploring if adding a compatibility directory that provides headers for exactly that makes things more understandable.

That make sense, but for now the volume of the OS-specific code is quite small. I think we can avoid creation of an additional abstraction layer now.

tests/integration/inverted/build.cpp
ibhati: Yeah, maybe just change it to 0.01 for both the cases.

tests/svs/lib/timing.cpp
```diff
@@ -45,6 +45,22 @@ void stress(svs::lib::Timer& timer, size_t max_depth = 3, size_t max_flat = 3) {
...
 }
+
+/*
+ * We use a busy_sleep here, not std::this_thread::sleep_for
+ * std::this_thread::sleep_for garanties, that the sleep would be
+ * not shorter than requared.
+ * For some platforms (like MacOS) a sleep can be much longer,
+ * so if we need the sleep time to be close to the requared,
+ * busy sleep is a good choise.
+ */
```
ibhati, line 56:
```suggestion
/*
 * We use a busy_sleep instead of std::this_thread::sleep_for.
 * std::this_thread::sleep_for guarantees that the sleep will be
 * not shorter than the required time.
 * For some platforms (like MacOS), a sleep can be much longer,
 * so if we need the sleep time to be close to the required,
 * busy sleep is a good choice.
 */
```
Some typos here. Also can you run clang-formating with these commands in the root directory:
sudo apt install clang-format-15 (if clang-format-15 not installed)
bash tools/clang-format.sh clang-format-15

.github/workflows/build-macos.yaml
```diff
@@ -0,0 +1,85 @@
...
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: ${{ matrix.cxx }}, ${{ matrix.build_type }}
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        build_type: [RelWithDebugInfo]
+        cxx: [clang++-15]
```
ibhati, line 38: Can we include at least one gcc compilation as well (maybe gcc-11)

include/svs/core/allocator.h
```diff
@@ -41,7 +41,16 @@
 #include <array>
 #include <fcntl.h>
 #include <filesystem>
+
+/* <linux/mman.h> provides some linux-specific flags like
+ * MAP_POPULATE, MAP_NORESERVE, MAP_HUGETLB.
+ * They can be used only with linux, and are not availiable
+ * for MacOS
+ */
```
ibhati, line 49: Just a nitpick for the sake of consistency: can we change the comments with `//` instead of `/*`?  This applies to other comments in the PR

ibhati (changes requested):

tests/integration/inverted/build.cpp
razdoburdin: > Yeah, maybe just change it to 0.01 for both the cases.

done

.github/workflows/build-macos.yaml
razdoburdin: Not with the current version :( GCC requires some additional tuning.

napetrov: I would say we should wait for GCC on mac until specific request. This would be quite a non standard setup.  Clang considered more native approach there.

Perf libs was supporting only clang on mac, back in the days

ibhati: Ok, sounds good to me

ibhati (approved):

ahuber21 (approved):
Thanks for addressing the comments. Let's keep the abstraction layer in mind. Many things in the library are implemented with future generalization in mind. It definitely wouldn't be out of place. So maybe experiment with it the next time you add support for new build infra or hardware! :)

napetrov (approved):

PR #112, opened 2025-04-18
Solve invalid memory read issue in find_nearest

dian-lun-lin:
This PR addresses the corner cases where neighbor id may be [max](https://github.com/intel/ScalableVectorSearch/blob/72633231807d46ff59f651916858fa0ed6ba3ee4/include/svs/lib/neighbor.h#L160) when distance is infinity. In such scenarios, the system could return a neighbor with the maximum ID, leading to an invalid memory read. To resolve this, the PR modifies the behavior to reset the ID to the first index, ensuring a valid existing neighbor is returned.


ibhati (approved):

aguerreb (approved):
Looks good to me.

PR #117, opened 2025-04-29
Update eve to the last version

razdoburdin:
This PR updates eve to the last version.
Unfortunately the last stable release was long time ago, and the problem with GCC and ARM isn't solved in a stable release. So I updated eve to the last commit in the main branch.


ahuber21:
Thanks for looking into this, it made it obvious that I was using `eve` incorrectly. After removal of `eve::reduce()` in my PR, the update is no longer necessary.

PR #119, opened 2025-04-30
Solve compression issue on ARM

dian-lun-lin:


dian-lun-lin:
The issue has been resolved by utilizing a blocked allocator, and I believe this PR is ready for merging. We could create a separate issue to address compression (as well as batch iterator) with a non-blocked allocator on the ARM platform.

ibhati (approved):

PR #122, opened 2025-05-12
fix: add logger everywhere we need to print

yuejiaointel:


include/svs/index/vamana/calibrate.h
```diff
@@ -346,8 +348,9 @@ VamanaSearchParameters tune_prefetch(
     Index& index,
     VamanaSearchParameters search_parameters,
     const DoSearch& do_search,
-    svs::logging::logger_ptr logger = svs::logging::get()
+    svs::logging::logger_ptr logger
 ) {
+    assert(logger != svs::logging::get());
```
dian-lun-lin, line 353: It seems unnecessary to perform this assertion. What if we want to pass a global logger?

include/svs/index/vamana/calibrate.h
```diff
@@ -177,8 +177,9 @@ VamanaSearchParameters optimize_split_buffer(
     VamanaSearchParameters current,
     const F& compute_recall,
     const DoSearch& do_search,
-    svs::logging::logger_ptr logger = svs::logging::get()
+    svs::logging::logger_ptr logger
 ) {
+    assert(logger != svs::logging::get());
```
dian-lun-lin, line 182: Same here.

include/svs/index/vamana/dynamic_index.h
```diff
@@ -936,10 +936,10 @@ class MutableVamanaIndex {
         auto entry_point = entry_point_[0];
         if (status_.at(entry_point) == SlotMetadata::Deleted) {
             auto logger = svs::logging::get();
```
dian-lun-lin, line 938: This logger should be removed. Not sure why, but our CI tests should fail to compile if a variable is unused.

include/svs/index/vamana/calibrate.h
yuejiaointel: Good catch, fixed! I was using it to test if there is any places I missed that needed custom logger

include/svs/index/vamana/calibrate.h
yuejiaointel: Thx, removed!

tests/svs/index/vamana/dynamic_index_2.cpp
```diff
@@ -445,6 +446,16 @@ CATCH_TEST_CASE("Dynamic MutableVamanaIndex Per-Index Logging Test", "[logging]"
...
+            global_captured_logs.emplace_back(msg.payload.data(), msg.payload.size());
+        }
+    );
+    global_callback_sink->set_level(spdlog::level::trace);
+
+    auto original_logger = svs::logging::get();
+    original_logger->sinks().push_back(global_callback_sink);
+
     // Setup index
     std::vector<float> data = {1.0f, 2.0f};
```
ibhati, line 460: Can we use the whole test dataset instead of a couple of them, because some of the build logs get printed after some batch of vectors are done?

tests/svs/index/vamana/index.cpp
```diff
@@ -141,6 +142,16 @@ CATCH_TEST_CASE("Static VamanaIndex Per-Index Logging", "[logging]") {
...
+            global_captured_logs.emplace_back(msg.payload.data(), msg.payload.size());
+        }
+    );
+    global_callback_sink->set_level(spdlog::level::trace);
+
+    auto original_logger = svs::logging::get();
+    original_logger->sinks().push_back(global_callback_sink);
+
     // Create some minimal data
     std::vector<float> data = {1.0f, 2.0f};
```
ibhati, line 156: Same here, use all the test dataset. Also, can we use `experimental_calibrate` and `search` after loading/building the index?

tests/svs/index/vamana/dynamic_index_2.cpp
```diff
@@ -462,6 +473,7 @@ CATCH_TEST_CASE("Dynamic MutableVamanaIndex Per-Index Logging Test", "[logging]"
     );
 
     // Verify the internal log messages
```
ibhati, line 475: Can we also use `add_vector` and `delete_vector`, `consolidate` and `compact` in this tests? I think the idea scenario is to incorporate your logging in the main test here [https://github.com/intel/ScalableVectorSearch/blob/main/tests/svs/index/vamana/dynamic_index_2.cpp#L251](url)
rather than recreate separate tests for logging

tests/svs/index/vamana/dynamic_index_2.cpp
yuejiaointel: Thx, replaced with test_dataset

tests/svs/index/vamana/index.cpp
yuejiaointel: Thx, replaced with test_dataset. But experimental_calibrate  and search are on orchestrator levels, now logging is only on low index levels. Will make another PR to make logging into orchestrator. But can we merge this one to unblock Rafik?

tests/svs/index/vamana/dynamic_index_2.cpp
yuejiaointel: Thx I integrated logging to the main test in dynamic_index_2. But similarly to the last comment for add_vector and delete_vector, consolidate, and compact, they are on orchestrator levels. Can we make merge this one first?

ibhati (approved):

PR #123, opened 2025-05-13
Add mutli-vector index and batch iterator support

dian-lun-lin:
This PR introduces multi-vector indexing and batch iterator functionality. With this PR, each label can be associated with multiple vectors, while ensuring that neighbors are retrieved with unique labels during search.

- Key APIs for `MultiBatchIterator`:
`next`: Retrieve a batch of neighbors, ensuring each neighbor has a unique label.

- Key APIs for `MultiMutableVamanaIndex`:
`search`: Utilize `MultiBatchIterator` to perform searches.
`add_points`: Allow the addition of points to the index, associating each with a corresponding label. Points with the same label can be added without issue.
`delete_entries`: Remove all vectors associated with a specific label from the index.
`get_distance`: Return the shortest distance among vectors with the given label and query.
`make_batch_iterator`: Return a `MultiBatchIterator` associated with the current index.

- Add `make_batch_iterator` to `MutableVamanaIndex` and `VamanaIndex` as well to enhance functionality.

#TODOs:

- [x] Add more unittests
- [x] Add different types of distance into unittests
- [ ] Documentation
- [ ] (TBD) Disable translator in `MutableVamana` for `MultiMutableVamanaIndex`
- [ ]  Elevate `MultiMutableVamana` at orchestrator/type-ersure level and eventually to python bindings

### DIFF
--- a/include/svs/index/vamana/multi.h
+++ b/include/svs/index/vamana/multi.h
@@ -58,24 +58,22 @@ template <typename Index, typename QueryType> class MultiBatchIterator {
     ) {
         const auto& external_to_label = index_.get_external_to_label_lookup();
         auto results_copy = results_;
-        auto extra_results_copy = extra_results_;
         results_.clear();
         get_results_from_extra(batch_size);
 
         while (results_.size() < batch_size && !batch_iterator_.done()) {
             try {
-                batch_iterator_.next(batch_size + 10, cancel);
+                batch_iterator_.next(batch_size, cancel);
             } catch (const ANNException& error) {
                 results_ = std::move(results_copy);
-                extra_results_ = std::move(extra_results_copy);
                 throw error;
             }
             for (auto& result : batch_iterator_) {
                 auto label = external_to_label.at(result.id());
 
                 if (returned_.find(label) == returned_.end()) {
-                    returned_.insert(label);
                     if (results_.size() < batch_size) {
+                        returned_.insert(label);
                         results_.push_back(Neighbor<label_type>{label, result.distance()});
                     } else {
                         extra_results_.push_back(Neighbor<label_type>{
@@ -117,12 +115,12 @@ template <typename Index, typename QueryType> class MultiBatchIterator {
         std::sort(extra_results_.rbegin(), extra_results_.rend(), TotalOrder(compare{}));
 
         while (results_.size() < batch_size && !extra_results_.empty()) {
-            auto top = extra_results_.front();
+            auto best = extra_results_.back();
             extra_results_.pop_back();
 
-            if (returned_.find(top.id()) == returned_.end()) {
-                returned_.insert(top.id());
-                results_.push_back(std::move(top));
+            if (returned_.find(best.id()) == returned_.end()) {
+                returned_.insert(best.id());
+                results_.push_back(std::move(best));
             }
         }
 
--- a/include/svs/index/vamana/multi.h
+++ b/include/svs/index/vamana/multi.h
@@ -35,17 +35,19 @@ template <typename Index, typename QueryType> class MultiBatchIterator {
 
     // Private type aliases
     using result_buffer_type = std::vector<value_type>;
-    /// Random-access iterator to `value_type` over the current batch of results.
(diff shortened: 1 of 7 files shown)


dian-lun-lin:
To expedite development, this PR will not include documentation or the functionality to disable the translator. Separate issues will be created to address these tasks.

dian-lun-lin:
/intelci
set ref=multi-vector

include/svs/index/vamana/multi.h
```diff
@@ -0,0 +1,563 @@
...
+
+                if (returned_.find(label) == returned_.end()) {
+                    if (results_.size() < batch_size) {
+                        returned_.insert(label);
+                        results_.push_back(Neighbor<label_type>{label, result.distance()});
+                    } else {
+                        extra_results_.push_back(Neighbor<label_type>{
+                            label, result.distance()});
+                    }
+                }
```
ibhati, line 83: Can't we check in the `else` condition if the distance corresponding to the already `returned_` label is more(for L2)/less(for MIP) than the distance corresponding to this node, then replace the old `result_` for the label with this one?

include/svs/index/vamana/multi.h
```diff
@@ -0,0 +1,563 @@
...
+            label_to_external_.at(entry_point).at(0),
+            distance_,
+            std::move(adds),
+            std::move(threadpool_proto),
+            std::move(logger)
+        );
+    }
+
+    // change translator to label -> external_id
+    // create a transformed translator where external_id == internal_id
```
ibhati, line 231:
```suggestion
    // Change translator to label -> external_id
    // Create a transformed translator where external_id == internal_id
```

Should these comments be below where this logic is implemented? Here we can have a short comment on what this function is doing

include/svs/index/vamana/multi.h
```diff
@@ -0,0 +1,563 @@
...
+            std::move(logger)
+        );
+    }
+
+    const auto& get_label_to_external_lookup() const { return label_to_external_; }
+    const auto& get_external_to_label_lookup() const { return external_to_label_; }
+    const ParentIndex& get_parent_index() const { return *index_; }
+
+    template <typename Query>
+    double get_distance(label_type label, const Query& query) const {
```
ibhati, line 274: Do we have any test for get_distance that this really works?

ibhati (commented):
Added few small comments. Overall looks good. 
Just so that we don't forget, one of the task we need to do is to elevate this MultiMutableVamana at orchestrator/type-ersure level and eventually to python bindings. But for the first PR, I agree this is suffiecient

tests/svs/index/vamana/multi.cpp
```diff
@@ -0,0 +1,221 @@
...
+            test_indices.size() * (num_duplicated + 1)
+        );
+
+        std::iota(test_indices.begin(), test_indices.end(), 0);
+        test_index.delete_entries(test_indices);
+        CATCH_REQUIRE(test_index.size() == num_duplicated);
+        CATCH_REQUIRE(
+            test_index.get_parent_index().size() ==
+            (num_duplicated * (num_duplicated + 1)) / 2
+        );
```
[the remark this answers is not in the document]
dian-lun-lin: Added comments. For `num_duplicated`, it would be too slow to duplicate the test dataset with larger number.

include/svs/index/vamana/multi.h
dian-lun-lin: Revised the code to replace old result.

include/svs/index/vamana/multi.h
dian-lun-lin: Yes, it's under `Step grouping` test section.

ibhati (approved):

tests/svs/index/vamana/multi.cpp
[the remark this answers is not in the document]
ibhati: Ok, makes sense. Thanks

dian-lun-lin:
/intelci
set ref=multi-vector

PR #126, opened 2025-05-19
Revert EVE to earlier version as gcc 11.2.0 gives compilation errors

ibhati:


dian-lun-lin (approved):

jfalcou:
Hi, EVE author here. Do you have any trace of the gcc issue? 
We can probably have a look at why it failed.

Best regards

ahuber21:
Hi @jfalcou, thanks for reaching out. By simply reverting this PR I wasn't able to reproduce the issue, so I'm not able to share more details. Hopefully @dian-lun-lin or @ibhati can help out.

ibhati:
> OK, from looking at the error log, I only see errors related to the use of std::hash. Am I missing something?

My bad. I updated my above comment with the correct error file and instructions to reproduce it in SVS

PR #130, opened 2025-05-22
Mark iterator done when no more new neighbors can be retreived,

ibhati:
Mark the iterator done when no more new neighbors can be retrieved, even though there may be more nodes available in the index.


dian-lun-lin:
Does this mean the graph can be disconnected?

ibhati:
> Does this mean the graph can be disconnected?

Yes or based on how large is the configured extra_buffer_capacity, some nodes that are at a long distance from the query and sparsely connected may not be retrieved.

dian-lun-lin (approved):

PR #131, opened 2025-05-22
Fix multi-vector batch iterator done() issue

dian-lun-lin:
With this PR, we can immediately detect if all labels are found without exhaustively checking all vectors.


rfsaliev (approved):
LGFM

ibhati (approved):

PR #133, opened 2025-05-29
New ISA dispatcher implementation

dian-lun-lin:
This ISA dispatcher allows detecting whether the running machine supports none AVX, AVX2, or AVX512, and it dispatches to the appropriate and optimized code path accordingly.

#TODOs:
1. ~ADD VNNI runtime detection support~
2. ~cmake with different AVX availability~
3. ~Python binding~
4. ~Validate and benchmarking~

### DIFF
--- a/include/svs/core/distance/distance_core.h
+++ b/include/svs/core/distance/distance_core.h
@@ -39,16 +39,6 @@ template <size_t N> constexpr bool is_dim_supported() {
     return false;
 }
 
-// template <size_t N>
-// constexpr size_t update_dim() {
-// if constexpr(is_dim_supported<N>()) {
-// return N;
-//}
-// else {
-// return Dynamic;
-//}
-//}
-
 using default_accum_type = float;
 
 template <Arithmetic Accum, typename T, size_t Extent>


cmake/multi-arch.cmake
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set(SVS_X86_SRC_DIR "${PROJECT_SOURCE_DIR}/include/svs/multi-arch/x86")
+set(SVS_X86
+  "${SVS_X86_SRC_DIR}/avx2.cpp,avx2,haswell"
+  "${SVS_X86_SRC_DIR}/avx512.cpp,avx512,cascadelake"
```
Alexsandruss, line 18: General AVX512 support starts from `skylake-avx512`.
Since the only difference between `skylake-avx512` and `cascadelake` is AVX512-VNNI, all AVX512 code branches of avx512.cpp might be compiled with VNNI instructions in theory, however this is unlikely in practice.

include/svs/core/distance/distance_core.h
```diff
@@ -26,6 +26,29 @@
...
+// template <size_t N>
+// constexpr size_t update_dim() {
+// if constexpr(is_dim_supported<N>()) {
+// return N;
+//}
+// else {
+// return Dynamic;
+//}
+//}
+
```
ibhati: Can we remove this commented code?

cmake/multi-arch.cmake
dian-lun-lin: Yes, it will compile with VNNI extension support. We have VNNI runtime check to see if the architecture supports VNNI. If the architecture does not support VNNI, we fallback to AVX2.

napetrov: > we fallback to AVX2.

shouldn't we just run AVX512 with no VNNI in this case?

dian-lun-lin: We don't have AVX512 implementation for int8/uint8, that's why currently I fallback to AVX2.

One interesting point is that the original main branch will fallback to non-AVX code, so we actually can gain some performance with this ISA dispatch ability on non AVX512VNNi architecture:)

PR #137, opened 2025-06-10
Batch iterator: clear the buffer when initializing

ibhati:
This is important when the buffer is repurposed for update() call


ibhati:
\intelci

PR #146, opened 2025-06-23
fix: bring logging level up and add addtional logging

yuejiaointel:
Bring logger up in level and add additional logging to print parameters


yuejiaointel:
> Thanks for the efforts! Could you add some tests to see if added logs are properly triggered at the correct level?

Hi Aaron, added level test in dynamic_index_2.cpp and index_build.cpp

dian-lun-lin (approved):

PR #152, opened 2025-06-27
Fix graph building with SQdata

ibhati:
During graph construction, both the query and dataset vectors are compressed. However, this compression was not handled correctly in the computation pipeline, leading to significantly reduced recall in some cases—especially when using normalized data with MIP (Maximum Inner Product) distance.


dian-lun-lin (approved):
LGTM

PR #159, opened 2025-07-21
Update to 0.0.8

ethanglaser:
Updates needed for 0.0.8 release

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,5 +19,3 @@
   * Static library provided in addition to .so
   
   * Intel(R) MKL linked within the shared library - no need for Intel(R) MKL in user environment
-
-* Various fixes and support regarding [RediSearch VectorSimilarity](https://github.com/RedisAI/VectorSimilarity) integration efforts
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Additions and Changes
 
-* Addition of scalar quantization dataset with fp16 support
+* Addition of scalar quantization dataset with int8 support
 
 * Introduced multi-vector index and batch iterator support
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,4 +2,22 @@
 
 ## Additions and Changes
 
-TODO
+* Addition of scalar quantization dataset with fp16 support
+
+* Introduced multi-vector index and batch iterator support
+
+* Automatic ISA dispatching with optimizations based on AVX support
+
+* Enhanced logging capabilities
+
+* Updated vamana iterator API
+
+* Broader [shared library](https://github.com/intel/ScalableVectorSearch/releases) support:
+
+  * gcc-11+, clang-18+, glibc 2.26+ compatibility
+  
+  * Static library provided in addition to .so
+  
+  * Intel(R) MKL linked within the shared library - no need for Intel(R) MKL in user environment
+
+* Various fixes and support regarding [RediSearch VectorSimilarity](https://github.com/RedisAI/VectorSimilarity) integration efforts
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Additions and Changes
 
-* Addition of scalar quantization dataset with int8 support
+* Addition of 8-bit scalar quantization support to C++ interface
 
 * Introduced multi-vector index and batch iterator support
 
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -1,13 +1,23 @@
...
+
+* Broader [shared library](https://github.com/intel/ScalableVectorSearch/releases) support:
+
+  * gcc-11+, clang-18+, glibc 2.26+ compatibility
+  
+  * Static library provided in addition to .so
+  
+  * Intel(R) MKL linked within the shared library - no need for Intel(R) MKL in user environment
+
+* Various fixes and support regarding [RediSearch VectorSimilarity](https://github.com/RedisAI/VectorSimilarity) integration efforts
```
mihaic:
```suggestion
```

NEWS.md
```diff
@@ -1,13 +1,21 @@
-# SVS 0.0.7 Release Notes
+# SVS 0.0.8 Release Notes
 
 ## Additions and Changes
 
-* Implemented batch iterator support for hybrid search
+* Addition of scalar quantization dataset with fp16 support
```
mihaic:
```suggestion
* Addition of scalar quantization dataset with int8 support
```

mihaic (changes requested):

NEWS.md
```diff
@@ -1,13 +1,21 @@
-# SVS 0.0.7 Release Notes
+# SVS 0.0.8 Release Notes
 
 ## Additions and Changes
 
-* Implemented batch iterator support for hybrid search
+* Addition of scalar quantization dataset with int8 support
```
aguerreb: Regarding the SQDataset, should we include this as experimental or something of the sort? Because it still doesn't have Python bindings for example. Also, I'd modify the sentence to something like: "Addition of 8-bit scalar quantization support." Because strictly speaking this is different from the int8 support that we had already in the previous version.

NEWS.md
```diff
@@ -1,13 +1,21 @@
...
 
-* Added support for custom threading and memory allocation
+* Introduced multi-vector index and batch iterator support
```
aguerreb: Explain what multi-vector refers to, as the concept is quite broad and use differently in different contexts. Here I believe it is multiple vectors can be mapped to the same external id, is that correct @dian-lun-lin?

aguerreb (changes requested):

NEWS.md
ethanglaser: Agreed that more detail here would be useful, @dian-lun-lin can you suggest better wording since you know it best - can push directly to this branch or comment suggestion

NEWS.md
ethanglaser:
```suggestion
* Addition of 8-bit scalar quantization support to C++ interface
```

ethanglaser: How does it look now?

NEWS.md
```diff
@@ -1,13 +1,21 @@
...
 ## Additions and Changes
 
-* Implemented batch iterator support for hybrid search
+* Addition of 8-bit scalar quantization support to C++ interface
 
-* Added support for custom threading and memory allocation
+* Introduced multi-vector index and batch iterator support
 
-* Introduced a timeout feature for search calls
+* Automatic ISA dispatching with optimizations based on AVX support
```
ethanglaser, line 9:
```suggestion
* Automatic ISA dispatching with optimizations based on AVX support

* Enabled compatibility with ARM and MacOS
```

NEWS.md
aguerreb: Looks good to me!

NEWS.md
dian-lun-lin: Introduced multi-vector index and batch iterator support that allows multiple vectors to be mapped to the same external ID.

NEWS.md
```diff
@@ -1,13 +1,23 @@
-# SVS 0.0.7 Release Notes
+# SVS 0.0.8 Release Notes
 
 ## Additions and Changes
 
-* Implemented batch iterator support for hybrid search
+* Addition of 8-bit scalar quantization support to C++ interface
 
-* Added support for custom threading and memory allocation
+* Introduced multi-vector index and batch iterator support
```
ethanglaser:
```suggestion
* Introduced multi-vector index and batch iterator support that allows multiple vectors to be mapped to the same external ID
```

aguerreb (approved):
Looks good to me, thanks for the changes!

PR #160, opened 2025-07-23
feature: add dynamic support for flat index

yuejiaointel:
To help integration with [Faiss](https://github.com/facebookresearch/faiss/issues/4450), this PR adds dynamic data support for flat index by using a generic Data instead of ImmutableMemoryDataset concept, can pass simple data to allow resize


bindings/python/src/dynamic_flat.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+namespace py = pybind11;
+namespace svs::python::dynamic_flat {
+
+namespace {
+
+template <typename F> void for_standard_specializations(F&& f) {
+#define X(Q, T, N) f.template operator()<Q, T, N>()
+    // Pattern:
+    // QueryType, DataType, Dimensionality, Enable Building
```
ibhati:
```suggestion
    // QueryType, DataType, Dimensionality
```

bindings/python/src/dynamic_flat.cpp
```diff
@@ -0,0 +1,311 @@
...
+const char* COMPACT_DOCSTRING = R"(
+Remove any holes created in the data by renumbering internal IDs.
+Shrink the underlying data structures.
+Following ``consolidate``, this can potentially reduce the memory footprint of the index
+if a sufficient number of points were deleted.
+)";
+
+const char* DELETE_DOCSTRING = R"(
+Soft delete the IDs from the index. Soft deletion does not remove the IDs from the data,
+but prevents them from being returned from future searches.
```
ibhati: For the Flat index, there's no separate index structure—it's just the data itself. So I think it would be clearer to refer to it as data instead of index in this context.

Also, I think we should directly delete the vector from the data itself, and there's no need for a soft delete mechanism in this case.

include/svs/index/flat/dynamic_flat.h
```diff
@@ -0,0 +1,692 @@
...
+/// @tparam Dist The distance functor used to compare queries with the elements of the
+///     dataset.
+///
+/// A flat index implementation that supports dynamic insertion and deletion of vectors
+/// while maintaining exhaustive search capabilities.
+///
+template <typename Data, typename Dist> class DynamicFlatIndex {
+    static_assert(
+        data::ImmutableMemoryDataset<Data>, "Data must satisfy ImmutableMemoryDataset"
+    );
```
ibhati: Why is this required. Data should be Mutable in the dynamic case. Am I missing something?

include/svs/index/flat/dynamic_flat.h
```diff
@@ -0,0 +1,692 @@
...
+    auto external_ids() const {
+        std::vector<size_t> ids;
+        ids.reserve(size());
+        on_ids([&ids](size_t id) { ids.push_back(id); });
+        return ids;
+    }
+
+    /// @brief Add the points with the given external IDs to the dataset.
+    ///
+    /// When `delete_entries` is called, a soft deletion is performed, marking the entries
```
ibhati: We should simplify the deletion as I mentioned above in case of Flat Index. I think we should directly delete the vector and make that Slot Empty.

include/svs/index/flat/dynamic_flat.h
```diff
@@ -0,0 +1,692 @@
...
+namespace svs::index::flat {
+
+///
+/// Metadata tracking the state of a particular data index.
+/// The following states have the given meaning for their corresponding slot:
+///
+/// * Valid: Valid and present in the associated dataset.
+/// * Deleted: Exists in the associated dataset, but should be considered as "deleted"
+/// and not returned from any search algorithms.
+/// * Empty: Non-existent and unreachable from standard entry points.
```
ibhati: There is no entry point in Flat Index. I think we need only two SlotMetadata in Flat: Empty and Valid. Does that make sense?

include/svs/index/flat/dynamic_flat.h
```diff
@@ -0,0 +1,692 @@
...
+        }
+    }
+
+    ///// Saving
+
+    /// @brief Save the index to disk.
+    void save(const std::filesystem::path& data_directory) const {
+        // Compact before saving to remove deleted entries
+        const_cast<DynamicFlatIndex*>(this)->compact();
+        lib::save_to_disk(data_, data_directory);
```
ibhati, line 404: I think we need to save Ids and translation as well, otherwise reloaded index will not work properly. Check out this code for Vamana https://github.com/intel/ScalableVectorSearch/blob/main/include/svs/index/vamana/dynamic_index.h#L1022

tests/svs/index/flat/dynamic_flat.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+            // Compact
+            tic = svs::lib::now();
+            // Use a batchsize smaller than the whole dataset to ensure that the compaction
+            // algorithm correctly handles this case.
+            index.compact(reference.valid() / 10);
+            diff = svs::lib::time_difference(tic);
+            do_check(index, reference, queries, diff, "compact");
+        }
+    }
```
ibhati, line 157: It would be good to add save/load test as well after the deletion and compaction step and see if the recall is good. Same for Python test

bindings/python/src/dynamic_flat.cpp
yuejiaointel: Thx, fixed!

bindings/python/src/dynamic_flat.cpp
yuejiaointel: Thx, changed!

include/svs/index/flat/dynamic_flat.h
yuejiaointel: Good catch, I think I just copied from flat, removed!

include/svs/index/flat/dynamic_flat.h
yuejiaointel: Thx, cleaned! just have valid or empty for status now

include/svs/index/flat/dynamic_flat.h
yuejiaointel: Thx, removed the deleted status

include/svs/index/flat/dynamic_flat.h
yuejiaointel: Yeh make sense, added config dir param in save and load

tests/svs/index/flat/dynamic_flat.cpp
yuejiaointel: Added the save and load test in test_dynamic_flat.py

PR #161, opened 2025-07-24
fix: add conitiona in shared test for testing

yuejiaointel:
Added an extra condition in shared test for testing different arch with sde


yuejiaointel:
/intelci
set ref=feature_add_arch_in_shared_lib_python_ci

PR #163, opened 2025-08-04
fix: add simple save

yuejiaointel:
Add save and load functions to flat index

### DIFF
--- a/include/svs/index/flat/flat.h
+++ b/include/svs/index/flat/flat.h
@@ -383,13 +383,14 @@ class FlatIndex {
         threads::parallel_for(
             threadpool_,
             threads::DynamicPartition{
-                queries.size(),
-                compute_query_batch_size(search_parameters, queries.size())},
+                queries.size(), compute_query_batch_size(search_parameters, queries.size())
+            },
             [&](const auto& query_indices, uint64_t /*tid*/) {
                 // Broadcast the distance functor so each thread can process all queries
                 // in its current batch.
                 distance::BroadcastDistance distances{
-                    extensions::distance(data_, distance_), query_indices.size()};
+                    extensions::distance(data_, distance_), query_indices.size()
+                };
 
                 search_patch(
                     queries,
@@ -569,7 +570,12 @@ template <data::ImmutableMemoryDataset Data, typename Dist, typename ThreadPoolP
 TemporaryFlatIndex<Data, Dist>
 temporary_flat_index(Data& data, Dist distance, ThreadPoolProto threadpool_proto) {
     return TemporaryFlatIndex<Data, Dist>{
-        data, distance, threads::as_threadpool(std::move(threadpool_proto))};
+        data, distance, threads::as_threadpool(std::move(threadpool_proto))
+    };
 }
+///// Saving and Loading
 
+void save(const std::filesystem::path& data_directory) const {
+    lib::save_to_disk(data_, data_directory);
+}
 } // namespace svs::index::flat


bindings/python/src/flat.cpp
```diff
@@ -223,6 +223,31 @@ void wrap(py::module& m) {
...
+        },
+        py::arg("data_directory"),
+        R"(
+Save the Flat index to disk.
+
+Args:
+    data_directory: Directory where the index data will be saved.
+)"
+    );
+
```
ibhati, line 247: To stay consistent with the Vamana implementation, could we wrap both assemble and save in separate functions with docstrings, and then call them inside wrap(..)? This would mirror the structure used here:

[assemble](https://github.com/intel/ScalableVectorSearch/blob/main/bindings/python/src/vamana.cpp#L322)
[save_index](https://github.com/intel/ScalableVectorSearch/blob/main/bindings/python/src/vamana.cpp#L280)

bindings/python/tests/test_flat.py
```diff
@@ -184,3 +184,32 @@ def test_from_array(self):
...
+        indices, distances = result
+        self.assertEqual(indices.shape, (1, 1))
+        self.assertEqual(indices[0, 0], 0)
+
+        # Clean up
+        if os.path.isdir(save_path):
+            import shutil
+            shutil.rmtree(save_path)
+        elif os.path.exists(save_path):
+            os.remove(save_path)
```
ibhati: I think it would be better to add the save/reload index test to the existing tests—perhaps in test_from_array. After each _do_test() call, we could save the index, reload it, and then run _do_test() again using the larger test dataset. It would also be good to cover different datatype combinations (fp32, fp16, int8, uint8). The [Vamana test](https://github.com/intel/ScalableVectorSearch/blob/main/bindings/python/tests/test_vamana.py#L249) can serve as a helpful reference for this.

include/svs/index/flat/flat.h
```diff
@@ -512,6 +513,12 @@ class FlatIndex {
         // Call extension for distance computation
         return svs::index::flat::extensions::get_distance_ext(data_, distance_, id, query);
     }
+
+    ///// Saving
+
+    void save(const std::filesystem::path& data_directory) const {
```
ibhati:
```suggestion

    ///// Saving
    void save(const std::filesystem::path& data_directory) const {
```

include/svs/index/flat/flat.h
```diff
@@ -569,7 +576,7 @@ template <data::ImmutableMemoryDataset Data, typename Dist, typename ThreadPoolP
 TemporaryFlatIndex<Data, Dist>
 temporary_flat_index(Data& data, Dist distance, ThreadPoolProto threadpool_proto) {
     return TemporaryFlatIndex<Data, Dist>{
-        data, distance, threads::as_threadpool(std::move(threadpool_proto))};
+        data, distance, threads::as_threadpool(std::move(threadpool_proto))
+    };
```
ibhati: Please run clang formatting before merge using this command:
``
bash tools/clang-format.sh clang-format-15
``

tests/svs/index/flat/flat.cpp
```diff
@@ -66,3 +67,38 @@ CATCH_TEST_CASE("FlatIndex Logging Test", "[logging]") {
...
+    //! [Loading]
+
+    // Verify loaded index by searching
+    std::vector<float> query{1.0f, 2.0f};
+    auto queryView = svs::data::SimpleDataView<float>(query.data(), 1, 2);
+    auto result = loaded_index.search(queryView, 1);
+
+    // Clean up
+    std::filesystem::remove_all(save_path);
+}
```
ibhati: I think it's better to tests save/load in the integration [test](https://github.com/intel/ScalableVectorSearch/blob/main/tests/integration/exhaustive.cpp) so that it covers different combinations with a larger test dataset. Use this vamana test as a [reference](https://github.com/intel/ScalableVectorSearch/blob/main/tests/integration/vamana/index_search.cpp#L297)

bindings/python/src/flat.cpp
yuejiaointel: Thx for the examples, fixed!

bindings/python/tests/test_flat.py
yuejiaointel: Thx for the examples, wrapped do test in save_reload_and_test so save and load will be tested for every case

include/svs/index/flat/flat.h
yuejiaointel: thx, fixed!

tests/svs/index/flat/flat.cpp
yuejiaointel: thx for the example, added in test flat function

bindings/python/src/flat.cpp
```diff
@@ -223,6 +227,27 @@ void wrap(py::module& m) {
...
+        &detail::save_index,
+        py::arg("data_directory"),
+        R"(
+Save a constructed Flat index to disk.
+
+Args:
+    data_directory: Directory where the dataset will be saved.
+
+Note: All directories should be separate to avoid accidental name collision with any
+auxiliary files that are needed when saving the various components of the index.
```
ibhati: I think we can remove this Note as there is only one directory in the Flat index unlike the Vamana index, which saves config/graph/data in multiple directories

bindings/python/tests/test_flat.py
```diff
@@ -151,29 +151,40 @@ def test_from_file(self):
...
+            if os.path.isdir(save_path):
+                shutil.rmtree(save_path)
+            elif os.path.exists(save_path):
+                os.remove(save_path)
 
         # Test `float32`
         print("Flat, From Array, Float32")
         flat = svs.Flat(data_f32, svs.DistanceType.L2)
-        self._do_test(flat, queries_f32, groundtruth, svs.DistanceType.L2, data_f32)
+        save_reload_and_test(flat, queries_f32, groundtruth, data_f32)
```
ibhati:
```suggestion
        self._do_test(flat, queries_f32, groundtruth, svs.DistanceType.L2, data_f32)
        save_reload_and_test(flat, queries_f32, groundtruth, data_f32)
```

ibhati: Please keep the original _do_test(...) call before doing save/load. The same comment applies to other calls below

tests/integration/exhaustive.cpp
```diff
@@ -77,6 +77,16 @@ void test_flat(
...
+    // Save and load
+    if constexpr (std::is_same_v<std::decay_t<Index>, svs::Flat>) {
+        svs_test::prepare_temp_directory();
+        auto temp_dir = svs_test::temp_directory();
+        index.save(temp_dir);
+        index = svs::Flat::assemble<svs::lib::Types<float, svs::Float16>>(
+            svs::VectorDataLoader<float>(temp_dir), distance_type, index.get_num_threads()
+        );
+    }
+
```
ibhati: Maybe do this in the end of this function followed by search and recall testing

bindings/python/src/flat.cpp
ibhati: Thanks!

bindings/python/src/flat.cpp
yuejiaointel: Thx, removed!

bindings/python/tests/test_flat.py
yuejiaointel: Thx, fixed by have save_and_load and a remove_dir called before and after _do_test

tests/integration/exhaustive.cpp
yuejiaointel: Thx, moved to end of the function!

ibhati (approved):
@yuejiaointel Thanks for incorporating the feedback. Now, it looks great to me!

PR #164, opened 2025-08-08
Handle corner in dynamic index with insufficient valid search results

ibhati:
This PR addresses a rare scenario in dynamic indexes where the search buffer may not be populated with enough results. This typically occurs when a large number of vectors have been deleted, resulting in a sparsely connected graph. To ensure robustness, the buffer is now supplemented with additional results when needed.

This is a corner case and is not expected to occur frequently, but handling it improves stability and consistency in edge conditions.

### DIFF
--- a/include/svs/index/vamana/extensions.h
+++ b/include/svs/index/vamana/extensions.h
@@ -434,14 +434,14 @@ template <typename Index, typename SearchBuffer, typename Query>
 void check_and_supplement_search_buffer(
     const Index& index, SearchBuffer& search_buffer, const Query& query
 ) {
-    if (search_buffer.valid() < search_buffer.target()) {
+    if (search_buffer.valid() < search_buffer.target_window() &&
+        search_buffer.valid() < index.size()) {
         for (auto external_id : index.external_ids()) {
             auto internal_id = index.translate_external_id(external_id);
             auto dist = index.get_distance(external_id, query);
             auto builder = index.internal_search_builder();
             search_buffer.insert(builder(internal_id, dist));
-
-            if (search_buffer.valid() >= search_buffer.target()) {
+            if (search_buffer.valid() >= search_buffer.target_window()) {
                 break;
             }
         }
--- a/include/svs/index/vamana/extensions.h
+++ b/include/svs/index/vamana/extensions.h
@@ -410,16 +410,18 @@ struct VamanaSingleSearchType {
         typename SearchBuffer,
         typename Scratch,
         typename Query,
-        typename Search>
+        typename Search,
+        typename Index>
     void operator()(
         const Data& data,
         SearchBuffer& search_buffer,
         Scratch& scratch,
         const Query& query,
         const Search& search,
+        const Index& index,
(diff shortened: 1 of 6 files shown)


include/svs/index/vamana/extensions.h
```diff
@@ -451,6 +478,10 @@ SVS_FORCE_INLINE void svs_invoke(
     // Perform graph search.
     auto accessor = data::GetDatumAccessor();
     search(query, accessor, distance, search_buffer);
+
+    if constexpr (Index::needs_id_translation) {
```
dian-lun-lin, line 482: Why do we only check the case that needs id translation?

include/svs/index/vamana/extensions.h
```diff
@@ -410,19 +410,44 @@ struct VamanaSingleSearchType {
...
+/// This can occur in dynamic indexes when many vectors have been deleted
+/// and the graph becomes sparsely connected. It's a corner case and should
+/// not happen frequently, but when it does, we may need to supplement the buffer
+/// with additional results.
+template <typename Index, typename SearchBuffer, typename Query>
+void check_and_supplement_search_buffer(
+    const Index& index, SearchBuffer& search_buffer, const Query& query
+) {
+    if (search_buffer.valid() < search_buffer.target_window() &&
+        search_buffer.valid() < index.size()) {
```
dian-lun-lin, line 438: Do we need to check this second condition? I thought the number of valid vectors should always be smaller than index size.

include/svs/index/vamana/extensions.h
ibhati: This check is specifically for dynamic indices and is necessary when a large number of vectors have been deleted from the original index. Ideally, we should implement a similar function for static indices as well, although such cases are extremely rare.

dian-lun-lin: I feel `supports_deletions` is more expressive as we are handling the case where many vectors are deleted. But this is minor.

ibhati: I know the name is a bit confusing, as it's used as a proxy to detect dynamic_index (see this line for example https://github.com/intel/ScalableVectorSearch/blob/main/include/svs/index/vamana/iterator.h#L191). But I think the overall quality of dynamic_index can be poor, since we add only a few vectors at a time and those nodes may not be fully connected. That’s why I’m applying this check for dynamic indices, even if they haven’t gone through deletion. This makes it safer—so if someone searches the index before a consolidation round, they get reasonable vectors instead of duplicates with index 0.

include/svs/index/vamana/extensions.h
ibhati: I was checking for a condition where target_window (i.e., search_window_size) is larger than index.size(), and valid() equals index.size(). In that case, I think the correct check should be:
`
search_buffer.valid() != index.size()
`
Does that make sense?

dian-lun-lin: I see your point now. We want to skip the insertion if all vectors are already in valid. In this case, we could keep the current checking as it is clearer. Thanks.

PR #170, opened 2025-08-13
Support omp threadpool

dian-lun-lin:
This pull request introduces a new OpenMP thread pool, `OMPThreadPool`, designed for users who wish to utilize OpenMP threading.


ibhati (commented):
To be consistent, can we follow the same pattern as SVS_ENABLE_NUMA for SVS_ENABLE_OMP as well? And that way also use it through SVS_VALIDATE_BOOL_ENV as done here https://github.com/intel/ScalableVectorSearch/blob/main/include/svs/lib/threads/threadpool.h#L117

PR #184, opened 2025-09-18
IVF follow-up to add compressed vector search

ibhati:


dian-lun-lin (approved):

bindings/python/tests/test_ivf.py
```diff
@@ -0,0 +1,308 @@
...
+                self.assertTrue(isapprox(recall, expected_recall, epsilon = 0.005))
+
+    def test_build(self):
+        # Build directly from data
+        queries = svs.read_vecs(test_queries)
+
+        # Build from file loader
+        loader = svs.VectorDataLoader(test_data_svs, svs.DataType.float32)
+        matcher = UncompressedMatcher("bfloat16")
+        self._test_build(loader, svs.DistanceType.L2, matcher)
```
ahuber21, line 307: I would have a dedicated test for each distance.

bindings/python/tests/test_ivf.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # Ensure the number of threads was propagated correctly.
+        self.assertEqual(ivf.num_threads, num_threads)
+
+        expected_results = self._get_config_and_recall(
+            'ivf_test_build', distance_map[distance], matcher
+        )
+
+        for expected in expected_results:
+            n_probes, k_reorder, k, nq, expected_recall = \
```
ahuber21, line 283: What is this auto-formatting with `\`? Is this configured in our formatter? That's ancient, we should update those rules.

include/svs/extensions/ivf/scalar.h
```diff
@@ -0,0 +1,48 @@
...
+}
+
+template <IsSQData Data, typename Alloc>
+auto svs_invoke(
+    svs::tag_t<index::ivf::extensions::create_dense_cluster>,
+    const Data& original,
+    size_t new_size,
+    const Alloc& SVS_UNUSED(allocator)
+) {
+    auto new_sqdata = SQDataset<typename Data::element_type, Data::extent>(
```
ahuber21, line 40: For my understanding: We create an empty SQDataset of `new_size` x "old" `dimensions` with same scale and bias?

include/svs/quantization/scalar/scalar.h
```diff
@@ -372,6 +372,9 @@ class SQDataset {
     using const_value_type = std::span<const element_type, Extent>;
     using value_type = const_value_type;
 
+    // Data wrapped in the library allocator.
+    using lib_alloc_data_type = SQDataset<T, Extent, lib::Allocator<T>>;
```
ahuber21, line 376: Is this used anywhere?

tests/integration/ivf/index_build.cpp
```diff
@@ -109,6 +109,7 @@ CATCH_TEST_CASE("IVF Build/Clustering", "[integration][build][ivf]") {
     test_build<svs::BFloat16>(svs::DistanceIP());
 
     // With 4 inner threads
-    test_build<svs::BFloat16>(svs::DistanceL2(), 4);
-    test_build<svs::BFloat16>(svs::DistanceIP(), 4);
+    // TBD: CI is not happy with this, investigate
```
ahuber21, line 112: Needs to be fixed before merge? Or follow-up?

ahuber21 (commented):
Posting as a comment without explicit approval or change request, as I will be offline for a few days soon.
Please address the points in my comments, then feel free to merge. Thanks!

tests/integration/ivf/index_build.cpp
ibhati: I will keep the comment for now, it passes on my local machine but not on the CI machines.

include/svs/quantization/scalar/scalar.h
ibhati: Yes, here https://github.com/intel/ScalableVectorSearch/blob/main/include/svs/index/ivf/index.h#L475

include/svs/extensions/ivf/scalar.h
ibhati: that's correct

bindings/python/tests/test_ivf.py
ibhati: We don't have formatting for Python code in the library

bindings/python/tests/test_ivf.py
ibhati: Yes, only L2 and MIP is supported in IVF right now.  Check here the corresponding Vamana tests: https://github.com/intel/ScalableVectorSearch/blob/main/bindings/python/tests/test_vamana.py#L377

PR #194, opened 2025-10-14
Rename global variable name to guard against conflicts

ibhati:
Renamed EPSILON to SVS_IVF_EPSILON to guard against conflicting names when SVS is used in bigger projects


mihaic:
VectorSimilarity is fixed and the fix is also backported to Redis 8.2:
https://github.com/RedisAI/VectorSimilarity/pull/799

PR #209, opened 2025-10-31
fix: store data members & make faiss tests pass

ahuber21:
Implementations should take care of storing their own data member values when saving. The `is_trained` logic for LeanVec had to be adapted to allow for use cases reported in https://github.com/ahuber21/faiss/issues/37

### DIFF
--- a/bindings/cpp/src/IndexSVSVamanaLeanVecImpl.cpp
+++ b/bindings/cpp/src/IndexSVSVamanaLeanVecImpl.cpp
@@ -104,22 +104,6 @@ svs::DynamicVamana* deserialize_impl_t(std::istream& stream, MetricType metric) 
 
 } // namespace
 
-IndexSVSVamanaLeanVecImpl* IndexSVSVamanaLeanVecImpl::build(
-    size_t dim,
-    MetricType metric,
-    const BuildParams& params,
-    size_t leanvec_dims,
-    LeanVecLevel leanvec_level
-) noexcept {
-    try {
-        auto index = new IndexSVSVamanaLeanVecImpl(
-            dim, params.graph_max_degree, metric, leanvec_dims, leanvec_level
-        );
-        index->build_params = params;
-        return index;
-    } catch (...) { return nullptr; }
-}
-
 IndexSVSVamanaLeanVecImpl::IndexSVSVamanaLeanVecImpl() = default;
 
 IndexSVSVamanaLeanVecImpl::IndexSVSVamanaLeanVecImpl(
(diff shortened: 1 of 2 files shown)


bindings/cpp/include/IndexSVSImplDefs.h
```diff
@@ -32,11 +32,12 @@ namespace runtime {
...
-    UNKNOWN_ERROR = 1,
-    INVALID_ARGUMENT = 2,
-    NOT_IMPLEMENTED = 3,
-    NOT_INITIALIZED = 4
+    SUCCESS,
+    UNKNOWN_ERROR,
+    INVALID_ARGUMENT,
+    NOT_IMPLEMENTED,
+    NOT_INITIALIZED,
+    IO_ERROR
```
rfsaliev: What is the reason to remove explicit constant values?
I used them to be 100% sure on ABI compatibility here

ahuber21: Ah I see the point. Do compilers actually behave differently?

bindings/cpp/src/IndexSVSVamanaImpl.cpp
```diff
@@ -466,12 +466,12 @@ Status IndexSVSVamanaImpl::init_impl(size_t n, const float* x) noexcept {
 }
 
 Status IndexSVSVamanaImpl::serialize_impl(std::ostream& out) const noexcept {
-    if (!impl) {
-        return Status{
-            ErrorCode::NOT_INITIALIZED, "Cannot serialize: SVS index not initialized."};
-    }
+    bool initialized = impl != nullptr;
```
rfsaliev, line 478: Can you please explain, why we have to serializae/deserialize empty inidices?

bindings/cpp/src/IndexSVSVamanaLeanVecImpl.cpp
```diff
@@ -237,13 +237,86 @@ Status IndexSVSVamanaLeanVecImpl::init_impl(size_t n, const float* x) noexcept {
...
+}
+
 Status IndexSVSVamanaLeanVecImpl::deserialize_impl(std::istream& in) noexcept {
     if (impl) {
         return Status{
             ErrorCode::INVALID_ARGUMENT,
             "Cannot deserialize: SVS index already initialized."};
     }
 
+    in.read(reinterpret_cast<char*>(&leanvec_d), sizeof(size_t));
```
rfsaliev: 1. I would reuse existing matrix serialization mechanizm implemented in SVS

bindings/cpp/src/IndexSVSVamanaLeanVecImpl.cpp
```diff
@@ -237,13 +237,86 @@ Status IndexSVSVamanaLeanVecImpl::init_impl(size_t n, const float* x) noexcept {
...
+        );
+
+        auto matrix = svs::data::SimpleData<float, svs::Dynamic>(num_rows, num_cols);
+        for (size_t i = 0; i < num_rows; ++i) {
+            const auto datum =
+                std::span<const float>(matrix_data.data() + i * num_cols, num_cols);
+            matrix.set_datum(i, datum);
+        }
+
+        leanvec_matrix =
```
rfsaliev: I would try to extract leanvec matrices from loaded `svs::leanvec::LeanDataset` instead of implementing custom serialization here.

bindings/cpp/src/IndexSVSVamanaLVQImpl.cpp
```diff
@@ -171,13 +171,31 @@ Status IndexSVSVamanaLVQImpl::init_impl(size_t n, const float* x) noexcept {
     );
 }
 
+Status IndexSVSVamanaLVQImpl::serialize_impl(std::ostream& out) const noexcept {
+    // Also store LVQ specific members
+    out.write(reinterpret_cast<const char*>(&lvq_level), sizeof(LVQLevel));
```
rfsaliev, line 176: So far, to make everything consistent, shouldn't we save `IndexSVSVamanaImpl` members as well?

bindings/cpp/src/IndexSVSVamanaImpl.cpp
ahuber21: For the use cases in this test
https://github.com/ahuber21/faiss/blob/b709fa114afc522b3d10ffd1356df1d9a9548951/tests/test_svs.cpp#L111

which is created to validate the scenario in this issue
https://github.com/ahuber21/faiss/issues/37

bindings/cpp/src/IndexSVSVamanaLeanVecImpl.cpp
ahuber21: Yes, now that it has been moved to SVS we can just use that. Is it difficult to re-use what you added for index save/load?

bindings/cpp/src/IndexSVSVamanaLeanVecImpl.cpp
ahuber21: In the current implementation it's possible to have a leanvec matrix (after running train) without having an index.

bindings/cpp/src/IndexSVSVamanaLVQImpl.cpp
ahuber21: ~~Isn't there only `ntotal_soft_deleted` which is always `0` because we compact on save?~~

Yes, I guess it makes sense to add the public members too.

bindings/cpp/src/IndexSVSVamanaImpl.cpp
rfsaliev: Thank you @ahuber21 for the problem description
After some time of investigation ana analysis, I got understanding, that inside the implementation code in SVS side we do not need to call `DynamicVamana::save(std::ostream&)` anymore, instead we can use filesystem-based `DynamicVamana::save(std::filesystem::path, ...)` method which allows to manage serialization in a more controlled and simpler way.

PR #223, opened 2025-11-18
Dynamic IVF Index Implementation

ibhati:
This PR enables support for **Dynamic IVF (Inverted File) Index** for SVS, enabling efficient vector search with dynamic insert and delete operations. The implementation includes comprehensive memory optimizations, thread configuration improvements, and full Python bindings.

### 🚀 Dynamic IVF Index
- **Dynamic Operations**: Support for adding and deleting vectors after index construction
- **Blocked Storage**: Uses `BlockedData` with configurable block sizes for efficient memory management
- **Thread Control**: Separate configuration for clustering threads, index threads, and intra-query threads

TODOs for IVF:
- Enable efficient Load/Save (for both static and dynamic IVF)
- Complete IVF documentation
- Batch iterator with IVF
- Benchmarking support for dynamic ivf

### DIFF
--- a/.github/workflows/build-cpp-runtime-bindings.yml
+++ b/.github/workflows/build-cpp-runtime-bindings.yml
@@ -61,7 +61,7 @@ jobs:
             /bin/bash -c "chmod +x docker/x86_64/build-cpp-runtime-bindings.sh && ./docker/x86_64/build-cpp-runtime-bindings.sh"
 
       - name: Upload cpp runtime bindings artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: svs-cpp-runtime-bindings${{ matrix.suffix }}
           path: svs-cpp-runtime-bindings${{ matrix.suffix }}.tar.gz
@@ -98,7 +98,7 @@ jobs:
 
       # Need to download for a new job
       - name: Download shared libraries
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v7
         with:
           name: svs-cpp-runtime-bindings${{ matrix.suffix }}
           path: runtime_lib
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ project(svs
     # - /bindings/python/tests/test_common.py
     # Manually keep in-sync with:
     # - /bindings/python/setup.py
-    VERSION 0.0.11
+    VERSION 0.1.0
 )
 
 set(SVS_LIB svs_devel)
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,13 @@
+# SVS 0.0.11 Release Notes
+
+## Additions and Changes
+
+* IVF support for SQDataset (#184), LVQ, and LeanVec
+
+* Added shared library support for gcc-14+
+
+* Various bug resolutions, fixes, and tooling support
+
 # SVS 0.0.10 Release Notes
 
 ## Additions and Changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,13 @@
-# SVS 0.0.11 Release Notes
+# SVS 0.1.0 Release Notes
 
 ## Additions and Changes
 
-IVF support for SQDataset (#184), LVQ, and LeanVec
+* Creation of SVS C++ Runtime Bindings package (#208)
 
-Various bug resolutions, fixes, and tooling support
+* Addition of `libsvs` conda and python packages along with existing tarballs
+
+* Added shared library support for Clang-20
+
+* LVQ8x8 support added to SVS python package
+
+* Fixes related to VamanaBuildParameters (#237) and unmasked AVX2 load (#239)
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 cmake_minimum_required(VERSION 3.21)
-project(svs_runtime VERSION 0.0.11 LANGUAGES CXX)
+project(svs_runtime VERSION 0.1.0 LANGUAGES CXX)
 set(TARGET_NAME svs_runtime)
 
 set(SVS_RUNTIME_HEADERS
--- a/bindings/cpp/include/svs/runtime/api_defs.h
+++ b/bindings/cpp/include/svs/runtime/api_defs.h
@@ -100,6 +100,7 @@ enum class StorageKind {
     FP16,
     SQI8,
     LVQ4x0,
+    LVQ8x0,
     LVQ4x4,
     LVQ4x8,
     LeanVec4x4,
--- a/bindings/cpp/src/svs_runtime_utils.h
+++ b/bindings/cpp/src/svs_runtime_utils.h
@@ -99,8 +99,8 @@ namespace storage {
 
 // Consolidated storage kind checks using constexpr functions
 inline constexpr bool is_lvq_storage(StorageKind kind) {
-    return kind == StorageKind::LVQ4x0 || kind == StorageKind::LVQ4x4 ||
-           kind == StorageKind::LVQ4x8;
+    return kind == StorageKind::LVQ4x0 || kind == StorageKind::LVQ8x0 ||
+           kind == StorageKind::LVQ4x4 || kind == StorageKind::LVQ4x8;
 }
 
 inline constexpr bool is_leanvec_storage(StorageKind kind) {
@@ -129,6 +129,7 @@ SVS_DEFINE_STORAGE_KIND_TAG(FP32);
 SVS_DEFINE_STORAGE_KIND_TAG(FP16);
 SVS_DEFINE_STORAGE_KIND_TAG(SQI8);
 SVS_DEFINE_STORAGE_KIND_TAG(LVQ4x0);
+SVS_DEFINE_STORAGE_KIND_TAG(LVQ8x0);
 SVS_DEFINE_STORAGE_KIND_TAG(LVQ4x4);
 SVS_DEFINE_STORAGE_KIND_TAG(LVQ4x8);
 SVS_DEFINE_STORAGE_KIND_TAG(LeanVec4x4);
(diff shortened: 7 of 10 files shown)


include/svs/index/ivf/common.h
```diff
@@ -224,56 +224,88 @@ template <typename T>
...
+    // Validate parameters to avoid Intel MKL errors
+    if (m == 0 || n == 0 || k == 0) {
+        return; // Nothing to compute
+    }
+
+    // Check for integer overflow when casting to int (MKL requirement)
+    constexpr size_t max_int = static_cast<size_t>(std::numeric_limits<int>::max());
+    if (m > max_int || n > max_int || k > max_int) {
+        throw ANNEXCEPTION(
+            "Matrix dimensions too large for Intel MKL GEMM: m={}, n={}, k={}", m, n, k
```
ahuber21: To make it more helpful we should include `max_int` in the error message.

include/svs/index/ivf/common.h
```diff
@@ -224,56 +224,88 @@ template <typename T>
...
+        );
+    }
+
     if constexpr (std::is_same_v<T, float>) {
+        // Cast size_t parameters to int for MKL GEMM functions
+        int m_int = static_cast<int>(m);
```
ahuber21: Do the casting outside the `if` block to avoid duplication

include/svs/index/ivf/index.h
```diff
@@ -164,6 +164,52 @@ class IVFIndex {
...
+        return id < id_to_cluster_.size() && id_to_cluster_[id] != SIZE_MAX;
+    }
+
+    ///// Distance Computation /////
+
+    /// @brief Compute the distance between a query vector and a vector in the index
+    template <typename Query> double get_distance(size_t id, const Query& query) const {
+        // Lazily initialize ID mapping on first call
+        if (id_to_cluster_.empty()) {
+            const_cast<IVFIndex*>(this)->initialize_id_mapping();
```
ahuber21: This is not thread-safe. I didn't find the pattern of `const_cast`ing `this` for lazy init elsewhere in the lib. So I suggest to not introduce it here.

include/svs/index/ivf/index.h
```diff
@@ -164,6 +164,52 @@ class IVFIndex {
...
+            );
+        }
+
+        // Get cluster and position
+        size_t cluster_id = id_to_cluster_[id];
+        size_t pos = id_in_cluster_[id];
+
+        // Fix distance argument if needed (e.g., for cosine similarity)
+        auto distance_copy = distance_;
+        svs::distance::maybe_fix_argument(distance_copy, query);
```
ahuber21, line 205:
```suggestion
        if constexpr (Dist::must_fix_argument) {
            // Fix distance argument if needed (e.g., for cosine similarity)
            auto distance_copy = distance_;
            svs::distance::maybe_fix_argument(distance_copy, query);
            // Call extension for distance computation
            return svs::index::ivf::extensions::get_distance_ext(
                cluster_, distance_copy, cluster_id, pos, query
            );
        } else {
            // Call extension for distance computation
            return svs::index::ivf::extensions::get_distance_ext(
                cluster_, distance_, cluster_id, pos, query
            );
        }
```

This might be premature optimization. Do you think we can gain a few QPS by avoiding the copy if it's not required?

Also: It states in `include/svs/index/ivf/common.h` that only L2 and MIP are supported. Is is helpful to mention cosine in the comment?

include/svs/index/ivf/index.h
ahuber21: Change to `constexpr`?
```suggestion
constexpr size_t MAX_QUERY_BATCH_SIZE = 10000;
```

include/svs/index/ivf/clustering.h
```diff
@@ -264,6 +270,12 @@ template <typename Data, std::integral I> struct DenseCluster {
 
     size_t size() const { return data_.size(); }
 
+    // Support for dynamic operations - SimpleData already has resize()
```
ahuber21:
```suggestion
    // Support for dynamic operations
```

`data_type` is templated at this point, we shouldn't mention `SimpleData`

include/svs/index/ivf/clustering.h
```diff
@@ -287,6 +299,7 @@ template <typename Data, std::integral I> struct DenseCluster {
     auto get_secondary(size_t id) const { return data_.get_secondary(id); }
     auto get_global_id(size_t local_id) const { return ids_[local_id]; }
     const Data& view_cluster() const { return data_; }
+    Data& view_cluster() { return data_; }
 
   public:
     Data data_;
```
ahuber21, line 305: Better change this to `data_type` for consistency? Some more examples of using `Data` instead of `data_type` exist throughout the file.

include/svs/index/ivf/dynamic_ivf.h
```diff
@@ -0,0 +1,973 @@
...
+        auto timer = lib::Timer();
+
+        // Process points in batches
+        for (size_t batch = 0; batch < num_batches; ++batch) {
+            auto batch_range = threads::UnitRange{
+                batch * batch_size, std::min((batch + 1) * batch_size, num_points)};
+
+            // Use centroid_assignment to compute assignments for this batch
+            centroid_assignment(
+                const_cast<Points&>(points), // centroid_assignment expects non-const
```
ahuber21: Again, I'm wary of using `const_cast`. I peaked into `centroid_assignment`. I don't see why it can't accept `const Data& data`. Can we change that?

(If the reason is that MKL doesn't accept const pointers, we should `const_cast` just before dispatching into MKL instead.)

include/svs/index/ivf/dynamic_ivf.h
```diff
@@ -0,0 +1,973 @@
...
+            "IDs size (", ids.size(), ") does not match data size (", data.size(), ")"
+        );
+    }
+
+    // Use lib_blocked_alloc_data_type for Dynamic IVF
+    using blocked_data_type = typename decltype(data)::lib_blocked_alloc_data_type;
+
+    // Use a small block size for IVF clusters (1MB instead of 1GB default)
+    auto blocking_params = data::BlockingParameters{
+        .blocksize_bytes = lib::PowerOfTwo(20) // 2^20 = 1MB
```
ahuber21, line 946: Why this value?

include/svs/index/ivf/kmeans.h
```diff
@@ -126,10 +143,11 @@ auto kmeans_clustering(
     Distance& distance,
     Pool& threadpool,
     lib::Type<I> integer_type = {},
-    svs::logging::logger_ptr logger = svs::logging::get()
+    svs::logging::logger_ptr logger = svs::logging::get(),
+    bool train_only = false
 ) {
     return kmeans_clustering_impl<BuildType>(
-        parameters, data, distance, threadpool, integer_type, std::move(logger)
+        parameters, data, distance, threadpool, integer_type, std::move(logger), train_only
```
ahuber21, line 150: Maybe split this into two functions? Along the lines of:

```c++
auto centroids = kmeans_centroids_impl<BuildType>(...);
std::vector<std::vector<uint32_t>> clusters(parameters.num_centroids_);
if (!train_only) {
  clusters = kmeans_clustering_impl<BuildType>(centroids, ...);
}
return std::make_tuple(centroids, std::move(clusters));
```

include/svs/index/ivf/index.h
ibhati: I don’t think this optimization is necessary for get_distance(). This function isn’t part of the main distance computation used in the build/search path—it’s primarily an external API for cases where someone wants to check the distance between a query and a specific vector.

include/svs/index/ivf/clustering.h
ibhati: Actually, the current usage is correct and idiomatic. The data_type alias is for external users of the class (like when someone writes DenseCluster<...>::data_type), while using Data directly inside the class is the normal C++ practice since it's the template parameter.

include/svs/index/ivf/dynamic_ivf.h
ibhati: There’s nothing special about this value. We can have thousands of clusters, and each should be initialized with the smallest possible block size to minimize extra memory usage. With the default block size and around 1000 clusters, the process was running out of memory. I tested with a 1 MB block size and didn’t observe any significant performance degradation when building dense clusters.

include/svs/index/ivf/kmeans.h
ibhati: I did not understand this comment. What is problem with the current method?

include/svs/index/ivf/common.h
ibhati: Done!

include/svs/index/ivf/common.h
ibhati: done!

include/svs/index/ivf/index.h
ibhati: Good catch. Fixed

include/svs/index/ivf/index.h
ibhati: Done

include/svs/index/ivf/clustering.h
ibhati: done

include/svs/index/ivf/dynamic_ivf.h
ibhati: Done!

include/svs/index/ivf/index.h
ahuber21: Agree

include/svs/index/ivf/clustering.h
ahuber21: Oh. That makes sense. Thanks

include/svs/index/ivf/kmeans.h
ahuber21: I didn't like the extra argument that was just slapped at the end of the impl and the relatively large `if/else` block. So this was just another nitpick about readability. Feel free to ignore if you see it differently.

include/svs/index/ivf/index.h
ahuber21: Could you clarify how the improved version is thread-safe?

ibhati: Now it should be thread safe, forgot to commit earlier ;)

include/svs/index/ivf/kmeans.h
ibhati: Good suggestion, but right now it's too much restructuring. But will keep this comment in mind for future improvements. thanks!

ahuber21 (approved):

include/svs/index/ivf/kmeans.h
ahuber21: Fine with me.

PR #227, opened 2025-11-20
Fix buffer invariants

ibhati:
After cleanup removes deleted nodes from the search buffer, its state becomes invalid. Call sort() to restore the buffer’s invariants and prepare it for the next round of iterator searches.


ahuber21:
What does it fix and how? Could you add a minimal description?

ahuber21 (approved):

PR #235, opened 2025-12-01
Make block_size flexible

razdoburdin:
This small PR adds an ability to set block_size value while building index. If parameter isn't set the default value of 2^30 is used.

### DIFF
--- a/docker/x86_64/build-cpp-runtime-bindings.sh
+++ b/docker/x86_64/build-cpp-runtime-bindings.sh
@@ -24,8 +24,7 @@ mkdir -p /workspace/bindings/cpp/build_cpp_bindings /workspace/install_cpp_bindi
 
 # Build and install runtime bindings library
 cd /workspace/bindings/cpp/build_cpp_bindings
-CC=gcc CXX=g++ cmake .. -S /workspace -B /workspace/bindings/cpp/build_cpp_bindings \
-  -DSVS_BUILD_RUNTIME_TESTS=ON -DCMAKE_INSTALL_PREFIX=/workspace/install_cpp_bindings -DCMAKE_INSTALL_LIBDIR=lib -DSVS_RUNTIME_ENABLE_LVQ_LEANVEC=${ENABLE_LVQ_LEANVEC:-ON}
+CC=gcc CXX=g++ cmake .. -DSVS_BUILD_RUNTIME_TESTS=ON -DCMAKE_INSTALL_PREFIX=/workspace/install_cpp_bindings -DCMAKE_INSTALL_LIBDIR=lib -DSVS_RUNTIME_ENABLE_LVQ_LEANVEC=${ENABLE_LVQ_LEANVEC:-ON}
 cmake --build . -j
 cmake --install .
 


bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -389,14 +390,17 @@ class DynamicVamanaIndexImpl {
         const index::vamana::VamanaBuildParameters& parameters,
         const svs::data::ConstSimpleDataView<float>& data,
         std::span<const size_t> labels,
+        int blocksize_exp,
         StorageArgs&&... storage_args
     ) {
         auto threadpool = default_threadpool();
 
+        auto blocksize_bytes = svs::lib::PowerOfTwo(blocksize_exp);
```
ahuber21: Since this is a runtime value, we'd probably need some validation. Do you think it could be a good idea to create a struct similar to `PowerOfTwo` and use that as function arg rather than the plain `int`. We could perform some boundary checks in the constructor.

bindings/cpp/tests/runtime_test.cpp
```diff
@@ -330,6 +331,40 @@ CATCH_TEST_CASE("LeanVecWithTrainingData", "[runtime]") {
...
+        CATCH_SKIP("Storage kind is not supported, skipping test.");
+    }
+    CATCH_REQUIRE(status.ok());
+    CATCH_REQUIRE(index != nullptr);
+
+    // Add data - should work with provided leanvec dims
+    std::vector<size_t> labels(test_n);
+    std::iota(labels.begin(), labels.end(), 0);
+
+    int block_size_exp = 17; // block_size = 2^block_size_exp
```
ahuber21: Beyond testing if passing the param works, we should also test if the block size actually changed.

bindings/cpp/include/svs/runtime/dynamic_vamana_index.h
```diff
@@ -29,7 +29,7 @@ namespace v0 {
 
 // Abstract interface for Dynamic Vamana-based indexes.
 struct SVS_RUNTIME_API DynamicVamanaIndex : public VamanaIndex {
-    virtual Status add(size_t n, const size_t* labels, const float* x) noexcept = 0;
+    virtual Status add(size_t n, const size_t* labels, const float* x, int blocksize_exp = 30) noexcept = 0;
```
ahuber21: I'm wondering if there's a cleaner way to communicate this value to `init_impl()`. Maybe @rfsaliev has an opinion?

bindings/cpp/include/svs/runtime/dynamic_vamana_index.h
```diff
@@ -29,7 +29,9 @@ namespace v0 {
 
 // Abstract interface for Dynamic Vamana-based indexes.
 struct SVS_RUNTIME_API DynamicVamanaIndex : public VamanaIndex {
-    virtual Status add(size_t n, const size_t* labels, const float* x) noexcept = 0;
+    virtual Status
+    add(size_t n, const size_t* labels, const float* x, int blocksize_exp = 30
```
rfsaliev: Seems like this approach manages block size for an index created from scratch.
How will we manage block size for an index loaded from a stream?

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -244,7 +252,11 @@ struct StorageFactory<SQStorageType> {
     using StorageType = SQStorageType;
 
     template <svs::threads::ThreadPool Pool>
-    static StorageType init(const svs::data::ConstSimpleDataView<float>& data, Pool& pool) {
+    static StorageType init(
+        const svs::data::ConstSimpleDataView<float>& data,
+        Pool& pool,
+        svs::lib::PowerOfTwo SVS_UNUSED(blocksize_bytes)
```
rfsaliev: `SQDataset<>::compress()` accepts customized allocator via `allocator` argument...

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -275,7 +287,11 @@ struct StorageFactory<LVQStorageType> {
     using StorageType = LVQStorageType;
 
     template <svs::threads::ThreadPool Pool>
-    static StorageType init(const svs::data::ConstSimpleDataView<float>& data, Pool& pool) {
+    static StorageType init(
+        const svs::data::ConstSimpleDataView<float>& data,
+        Pool& pool,
+        svs::lib::PowerOfTwo SVS_UNUSED(blocksize_bytes)
```
rfsaliev: `LVQDataset<>::compress()` accepts customized allocator via `allocator` argument...

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -309,6 +325,7 @@ struct StorageFactory<LeanVecStorageType> {
     static StorageType init(
         const svs::data::ConstSimpleDataView<float>& data,
         Pool& pool,
+        svs::lib::PowerOfTwo SVS_UNUSED(blocksize_bytes),
```
rfsaliev: `LeanDataset<>::compress()` accepts customized allocator via `allocator` argument...

bindings/cpp/src/dynamic_vamana_index_impl.h
razdoburdin: Yes, that make sense.

What values of exponents are acceptable? In the best of my understanding, blocksize_bytes should be in range [4KB, 1GB] (based on possible page sizes https://github.com/intel/ScalableVectorSearch/blob/d88aec87934952c935bbecb89aa1576dfcf2ed0e/include/svs/core/allocator.h#L95) , right?

ahuber21: Yes, that sounds reasonable. @mihaic did you ever experiment with even bigger hugepages?

mihaic: Since Xeon only goes to 1 GiB, we can have that as the limit.

bindings/cpp/tests/runtime_test.cpp
razdoburdin: I have added getter, to check the actual value of blocksize

bindings/cpp/src/dynamic_vamana_index_impl.h
razdoburdin: I have added check on object construction

bindings/cpp/src/svs_runtime_utils.h
razdoburdin: fixed

bindings/cpp/src/svs_runtime_utils.h
razdoburdin: fixed

bindings/cpp/src/svs_runtime_utils.h
razdoburdin: fixed

bindings/cpp/include/svs/runtime/dynamic_vamana_index.h
razdoburdin: `blocksize` is directly loaded from stream together with all the data, so for indexes being already saved the default `blocksize` would be used (as far as it has been already saved). 

I have added a check in runtime tests to verify that `blocksize` is writed and saved correctly.

ethanglaser:
@razdoburdin please re-open PR to main branch now that #208 was merged

PR #239, opened 2025-12-09
fix(simd): umasked AVX2 load

ahuber21:
The previous code _always_ performed a full width load on the provided data. In ragged-epilogue scenarios, where we request a masked load, this resulted in SEGV errors in certain runs with address sanitizer.

```c++
    if (i < count.size()) {
        auto mask = create_mask<simd_width>(count);
        s0 = op.accumulate(mask, s0, op.load_a(mask, a + i), op.load_b(mask, b + i));
    }
```

**Why wasn't this caught sooner?**

The OS only triggers a segmentation fault if a read accesses an unmapped memory page. Since memory protection (typically) operates at a 4KB page granularity, reading past the end of a buffer is "safe" from the OS's perspective unless the overflow happens to cross exactly into an unmapped page.

**Why is ASan catching it sporadically?**

Since our underlying object storage is `std::vector`, ASan detection requires two specific conditions to align:

* **No Spare Capacity:** The vector's `size()` must equal its `capacity()`. If there is spare capacity, the unsafe load simply reads valid (though uninitialized) memory owned by the vector.
* **Alignment & Redzones:** The underlying heap allocation must be sized and aligned such that the full-width SIMD read (e.g., 32 bytes) actually crosses the allocation boundary into the ASan redzone. If the allocator adds padding for alignment, the read might land in that valid padding instead.

### DIFF
--- a/include/svs/index/vamana/search_buffer.h
+++ b/include/svs/index/vamana/search_buffer.h
@@ -340,7 +340,7 @@ template <typename Idx, typename Cmp = std::less<>> class SearchBuffer {
     /// returns ``true``.
     ///
     bool can_skip(float distance) const {
-        return compare_(back().distance(), distance) && full();
+        return full() && (capacity() == 0 || compare_(back().distance(), distance));
     }
 
     ///


.github/workflows/asan.yml
```diff
@@ -0,0 +1,67 @@
...
+# and your use of them is governed by the express license under which they
+# were provided to you ("License"). Unless the License provides otherwise,
+# you may not use, modify, copy, publish, distribute, disclose or transmit
+# this software or the related documents without Intel's prior written
+# permission.
+#
+# This software and the related documents are provided as is, with no
+# express or implied warranties, other than those that are expressly stated
+# in the License.
+
```
ibhati: This should be open source /Apache license

.github/workflows/asan.yml
```diff
@@ -0,0 +1,67 @@
...
+name: Linux Build and Test
```
ibhati: Maybe change the name which will help us identify this is ASAN test

tests/svs/index/inverted/clustering.cpp
```diff
@@ -385,7 +385,7 @@ void test_end_to_end_clustering(
 
 } // namespace
 
-CATCH_TEST_CASE("Random Clustering - End to End", "[inverted][random_clustering]") {
+CATCH_TEST_CASE("Random Clustering - End to End", "[long][inverted][random_clustering]") {
```
ibhati, line 388: What is this [long]? Why is this required?

ibhati (approved):
The new workflow is added in the existing workflows, please make sure the earlier flow and options did not change.

.github/workflows/build-linux.yml
```diff
@@ -25,25 +25,43 @@ permissions:
 
 # This allows a subsequently queued workflow run to interrupt previous runs
 concurrency:
-  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
```
ethanglaser: In the other .github yml files all concurrency: group: lines use single quote - is there a reason for the change?

.github/workflows/build-linux.yml
```diff
@@ -69,25 +94,25 @@ jobs:
...
 
     - name: Build Tests and Utilities
       working-directory: ${{ runner.temp }}/build
       run: make -j$(nproc)
 
     - name: Run tests
       env:
-          CTEST_OUTPUT_ON_FAILURE: 1
+        CTEST_OUTPUT_ON_FAILURE: 1
+        ASAN_OPTIONS: detect_leaks=0
```
ethanglaser: false positives if this is 1?

.github/workflows/build-linux.yml
ahuber21: Sorry, was the auto-formatter. This will be made future-proof after merging
https://github.com/intel/ScalableVectorSearch/pull/240

.github/workflows/build-linux.yml
ahuber21: That was just some hallucination. But it doesn't need to be there I think. Let me try removing the option.

tests/svs/index/inverted/clustering.cpp
ahuber21: Sorry, had a comment that I didn't post. I added a new tag `[long]` that marks long-running tests. They are skipped in the debug asan build.

https://github.com/intel/ScalableVectorSearch/blob/63e58cd076c79d29b82dd2cd184de01044e7a5f9/.github/workflows/build-linux.yml#L63-L64

ahuber21:
> Do we know how long the ASan run takes if completing successfully?

@ethanglaser about 10 minutes total, < 5 mins of which is testing.

PR #240, opened 2025-12-10
Add yml linter

ahuber21:
I noticed inconsistent configuration of yml autoformatters, so I added a formatter to the pre-commit hooks.


ethanglaser (approved):
No objections from my end. Seems like an industry standard tool with fairly typical args. Thanks for unifying

PR #248, opened 2025-12-16
Make index blocksize flexible

razdoburdin:
This PR adds ability to set custom index blocksize 
Reopening of https://github.com/intel/ScalableVectorSearch/pull/235

The new parameter structure for build method of `DynamicVamanaIndex` and `DynamicVamanaIndexLeanVec` is introduced.
Example of building a LeanVec index with block_size = 2^12
```
svs::runtime::v0::DynamicVamanaIndex* index = nullptr;
svs::runtime::v0::VamanaIndex::BuildParams build_params{64};
svs::runtime::v0::VamanaIndex::DynamicIndexParams dynamic_index_params{12};
svs::runtime::v0::Status status = svs::runtime::v0::DynamicVamanaIndexLeanVec::build(
    &index,
    test_d,
    svs::runtime::v0::MetricType::L2,
    svs::runtime::v0::StorageKind::LeanVec4x4,
    32,
    build_params,
    {},
    dynamic_index_params
);
```

### DIFF
--- a/.github/workflows/build-cpp-runtime-bindings.yml
+++ b/.github/workflows/build-cpp-runtime-bindings.yml
@@ -68,7 +68,7 @@ jobs:
           retention-days: 7  # Reduce retention due to size
 
       - name: Upload conda package artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: libsvs-runtime-conda${{ matrix.suffix }}
           path: conda-bld/linux-64/libsvs-runtime-*.conda


rfsaliev:
Hi!
Thank you for the PR.
Looks good but I see some issues:
* Extra dependencies on SVS internal headers (`svs/lib/...`) added to the runtime API in `index_blocksize.h` - I would avoid such dependencies by replacing:
  *  usage `svs::lib::PowerOfTwo` with just a `size_t BlockSizeExponent();` and `size_t BlockSizeBytes() const { return 1 << blocksize_exp_; }`
  * replace `IndexBlockSize` .ctor with a static factory method e.g. `static Status IndexBlockSize::make(size_t blocksize_exp) noexcept;` - as it was made to other runtime API classes.
* The `DynamicVamanaIndex::add(..., IndexBlockSize);` method has unclear semantic, because `blocksize` argument can be applied at the first call only, but not at subsequential calls.

Instead I would propose to setup block size in a `build()` method:
* One of solutions can be blocksize field in `VamanaIndex::BuildParams` - but it seems block size makes no sense in non-dynamic indices.
* Another option: define a dedicated structure with parameters for dynamic indices: e.g. blocksize, reuse_empty_slots, etc.

bindings/cpp/include/svs/runtime/index_blocksize.h
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstddef>
+
+#include <svs/lib/exception.h>
+#include <svs/lib/misc.h>
```
rfsaliev: Please avoid SVS internals dependencies in SVS Runtime

bindings/cpp/include/svs/runtime/index_blocksize.h
```diff
@@ -0,0 +1,46 @@
...
+class IndexBlockSize {
+    constexpr static size_t kMaxBlockSizeExp = 30; // 1GB
+    constexpr static size_t kMinBlockSizeExp = 12; // 4KB
+
+    svs::lib::PowerOfTwo blocksize_bytes_;
+
+  public:
+    explicit IndexBlockSize(size_t blocksize_exp) {
+        if (blocksize_exp > kMaxBlockSizeExp) {
+            throw ANNEXCEPTION("Blocksize is too large!");
```
rfsaliev: SVS runtime API should be `noexcept`

bindings/cpp/src/dynamic_vamana_index.cpp
```diff
@@ -55,6 +55,16 @@ struct DynamicVamanaIndexManagerBase : public DynamicVamanaIndex {
     DynamicVamanaIndexManagerBase& operator=(DynamicVamanaIndexManagerBase&&) = default;
     ~DynamicVamanaIndexManagerBase() override = default;
 
+    Status
+    add(size_t n, const size_t* labels, const float* x, IndexBlockSize blocksize
+    ) noexcept override {
+        return runtime_error_wrapper([&] {
+            svs::data::ConstSimpleDataView<float> data{x, n, impl_->dimensions()};
+            std::span<const size_t> lbls(labels, n);
+            impl_->add(data, lbls, blocksize);
```
rfsaliev: Please convert SVS Runtime API data types (`IndexBlockSize`) to SVS internal type (`PowerOfTwo`) here.
As it was made for `data` argument

include/svs/index/vamana/dynamic_index.h
```diff
@@ -722,6 +728,16 @@ class MutableVamanaIndex {
         return slots;
     }
 
+    lib::PowerOfTwo blocksize_bytes() const {
```
rfsaliev: I do not think it makes sense to modify SVS internals just to implement a query for SVS runtime API

bindings/cpp/include/svs/runtime/index_blocksize.h
razdoburdin: done

bindings/cpp/include/svs/runtime/index_blocksize.h
razdoburdin: the `IndexBlockSize `structure is removed

include/svs/index/vamana/dynamic_index.h
razdoburdin: fixed
The getter now calls for parameter filed, not the SVS internals

bindings/cpp/src/dynamic_vamana_index.cpp
```diff
@@ -181,13 +204,20 @@ Status DynamicVamanaIndexLeanVec::build(
     StorageKind storage_kind,
     size_t leanvec_dims,
     const DynamicVamanaIndex::BuildParams& params,
-    const DynamicVamanaIndex::SearchParams& default_search_params
+    const DynamicVamanaIndex::SearchParams& default_search_params,
+    const DynamicVamanaIndex::DynamicIndexParams& dynamic_index_params
 ) noexcept {
     using Impl = DynamicVamanaIndexLeanVecImpl;
     *index = nullptr;
     return runtime_error_wrapper([&] {
```
rfsaliev, line 283: Does it make sense to call `DynamicVamanaIndex::check_params()` here as it is done in `DynamicVamanaIndex::build()`?

bindings/cpp/src/dynamic_vamana_index.cpp
```diff
@@ -201,15 +231,22 @@ Status DynamicVamanaIndexLeanVec::build(
     StorageKind storage_kind,
     const LeanVecTrainingData* training_data,
     const DynamicVamanaIndex::BuildParams& params,
-    const DynamicVamanaIndex::SearchParams& default_search_params
+    const DynamicVamanaIndex::SearchParams& default_search_params,
+    const DynamicVamanaIndex::DynamicIndexParams& dynamic_index_params
 ) noexcept {
     using Impl = DynamicVamanaIndexLeanVecImpl;
     *index = nullptr;
     return runtime_error_wrapper([&] {
```
rfsaliev, line 315: Same as above.

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -196,6 +197,27 @@ template <> struct StorageFactory<UnsupportedStorageType> {
...
+            pool,
+            svs::threads::StaticPartition(result.size()),
+            [&](auto is, auto SVS_UNUSED(tid)) {
+                for (auto i : is) {
+                    result.set_datum(i, data.get_datum(i));
+                }
+            }
+        );
+        return result;
+    }
```
rfsaliev, line 221: IMHO, would be better to modify the existing `init()` method rather than add an overload.
Is there any call to `StorageFactory<SimpleDatasetType<ElementType>>::init()` without `blocksize_bytes` argument?

razdoburdin: Yes, `FlatIndexImpl`  [calls for make_storage](https://github.com/intel/ScalableVectorSearch/blob/aa1f96d9e351fa675504a1be7d8d634ed4e78358/bindings/cpp/src/flat_index_impl.h#L121)

bindings/cpp/src/dynamic_vamana_index.cpp
razdoburdin: done

bindings/cpp/src/dynamic_vamana_index.cpp
razdoburdin: done

bindings/cpp/src/svs_runtime_utils.h
rfsaliev: What about applying 'default argument' value for `blocksize_bytes`?
For example:
```cpp
template <svs::threads::ThreadPool Pool>
    static StorageType init(
        const svs::data::ConstSimpleDataView<float>& data,
        Pool& pool,
        svs::lib::PowerOfTwo blocksize_bytes = svs::data::BlockingParameters::default_blocksize_bytes
    ) {
```
Alternative:
```cpp
template <svs::threads::ThreadPool Pool>
    static StorageType init(
        const svs::data::ConstSimpleDataView<float>& data,
        Pool& pool,
        svs::lib::PowerOfTwo blocksize_bytes = {0}
    ) {
        auto parameters = svs::data::BlockingParameters{.blocksize_bytes = blocksize_bytes};
        using alloc_type = typename StorageType::allocator_type;
        auto alloc = blocksize_bytes.raw() == 0 ? alloc_type{} :  alloc_type{parameters};
...
```

razdoburdin: done

rfsaliev (approved):
LGFM

PR #251, opened 2025-12-18
GitHub actions and general CI maintenance

ethanglaser:
Several CI related cleanup/refactoring tasks, functionality should largely be unaffected. So far:

- Replaces renovate with dependabot to align with private source repo, with monthly scans for actions updates
- Set all actions to the latest stable major version for consistency
- Migrate CI-related bash scripts from `docker/` to `.github/scripts/`
- Remove unnecessary chmod on above scripts
- Update cpp bindings and shared examples CMakeLists.txt to reference latest release binaries
- Throw error if ctests aren't found in shared/bindings validation
- Run Faiss-SVS examples for validation in addition to tests

### DIFF
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -98,14 +98,14 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
     else()
         # Links to LTO-enabled static library, requires GCC/G++ 11.2
         if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.3")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.1.0/svs-shared-library-0.1.0-lto.tar.gz"
+            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-lto-nightly-2026-02-05-1017.tar.gz"
                 CACHE STRING "URL to download SVS shared library")
         else()
             message(WARNING
                 "Pre-built LVQ/LeanVec SVS library requires GCC/G++ v.11.2 to apply LTO optimizations."
                 "Current compiler: ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION}"
             )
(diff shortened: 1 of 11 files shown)


.github/scripts/test-cpp-runtime-bindings.sh
```diff
@@ -46,10 +49,52 @@ echo "-----------------------------------------------"
...
 PYTHONPATH=../build/faiss/python/build/lib/ OMP_NUM_THREADS=8 python -m unittest test_svs_py.py
+echo "-----------------------------------------------"
+echo " FAISS-SVS python examples: "
+cd ../tutorial/python/
+if grep -q "GenuineIntel" /proc/cpuinfo; then
+  PYTHONPATH=../../build/faiss/python/build/lib/ OMP_NUM_THREADS=8 $RUN_PREFIX python 11-SVS.py
+else
+  echo "Non-Intel CPU detected - SVS python example expected to fail"
+  set +e
+  PYTHONPATH=../../build/faiss/python/build/lib/ OMP_NUM_THREADS=8 python 11-SVS.py
```
mihaic: I think we will be running on [4 hardware threads](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories).
```suggestion
  PYTHONPATH=../../build/faiss/python/build/lib/ OMP_NUM_THREADS=4 python 11-SVS.py
```

mihaic (changes requested):

.github/workflows/build-cpp-runtime-bindings.yml
```diff
@@ -58,17 +58,17 @@ jobs:
             -e ENABLE_LVQ_LEANVEC=${{ matrix.enable_lvq_leanvec }} \
             -e SUFFIX=${{ matrix.suffix }} \
             svs-manylinux228:latest \
-            /bin/bash -c "chmod +x docker/x86_64/build-cpp-runtime-bindings.sh && ./docker/x86_64/build-cpp-runtime-bindings.sh"
+            /bin/bash .github/scripts/build-cpp-runtime-bindings.sh
 
       - name: Upload cpp runtime bindings artifacts
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v4
```
ahuber21, line 64: Why the downgrade?

ethanglaser: Align with private codebase, which has issues with v6

ahuber21: Noticed that, thanks for clarification.

PR #262, opened 2026-01-27
feat(runtime): update to v1 API for leanvec ood

ahuber21:
Increments runtime API version to `v1`, because `LeanVecTrainingData::build()` changed significantly for OOD suport.

Unchanged entities from `v0` are aliased in `v1`.

Successful runtime lib build pending changes from svs prebuilt binaries.

### DIFF
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -98,14 +98,14 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
     else()
         # Links to LTO-enabled static library, requires GCC/G++ 11.2
         if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.3")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v1.0.0-dev/svs-shared-library-1.0.0-NIGHTLY-20251119-807-lto.tar.gz"
+            set(SVS_URL "<PLEASE_UPDATE_ME>"
                 CACHE STRING "URL to download SVS shared library")
         else()
             message(WARNING
                 "Pre-built LVQ/LeanVec SVS library requires GCC/G++ v.11.2 to apply LTO optimizations."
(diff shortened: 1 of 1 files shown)


bindings/cpp/include/svs/runtime/training.h
```diff
@@ -27,6 +27,15 @@ namespace v0 {
 
 struct SVS_RUNTIME_API LeanVecTrainingData {
     virtual ~LeanVecTrainingData();
+
+    /* Build LeanVec training data (compression matrices) from the provided
+     * data.
```
ibhati, line 32: Should we call "Leanvec transformation matrices" instead of training data?

bindings/cpp/include/svs/runtime/training.h
```diff
@@ -42,5 +51,37 @@ struct SVS_RUNTIME_API LeanVecTrainingData {
...
+     * @param q Pointer to the training queries (can be nullptr)
+     * @param leanvec_dims Number of dimensions in the resulting LeanVec data
+     */
+    static Status build(
+        LeanVecTrainingData** training_data,
+        size_t dim,
+        size_t n,
+        const float* x,
+        size_t n_train,
+        const float* q,
```
ibhati, line 78: Can we keep these two new arguments in the end and default initialize to 0 and nullptr? In that case, the older v0 calls to build will still work without any modifications. Also "n_train" is kind of confusing as both data/queries are used for training? How about we explicitly say "n_queries"?

ahuber21: That's exactly the discussion I wanted to have. Your suggestion would indeed allow us to stick with v0 (only that we'd need a copy of the function and we can't use default values, due to ABI compability, but that's just a detail).

In my opinion, the order `(n_data, const float* data, n_query, const float* queries, size_t leanvec_dims)` just makes more sense than `(n_data, const float* data, size_t leanvec_dims, n_query, const float* queries)`. But does it justify bumping to v0? 

Your preference would be to stick with `v0` for as long as possible?

bindings/cpp/include/svs/runtime/training.h
ahuber21: LeanVecTrainingData was purposefully chosen to be generic. In hindsight, I think we could have been more specific. But changing this would require an API update and therefore conflicts with your suggestion in the other comment.

PR #264, opened 2026-01-28
feat(runtime): add leanvec ood

ahuber21:
Add LeanVec OOD API to runtime.

Supersedes #262 as it achieves the same thing with only a minor version bump.


rfsaliev (approved):
LGFM

PR #276, opened 2026-02-17
docs: add hierarchical AGENTS and long-lived Copilot instructions

napetrov:
## Summary
- Add hierarchical AGENTS.md files across key repo areas for scoped agent context
- Add GitHub Copilot instructions that enforce AGENTS-first context loading
- Keep instruction docs long-lived and route mutable details to source-of-truth files

## Details
- Root AGENTS.md as compact entry-point/router
- Directory-level AGENTS.md for .github/, benchmark/, bindings/, cmake/, examples/, include/svs/, tests/, tools/, utils/
- .github/copilot-instructions.md with required flow: root → nearest → most specific
- .github/instructions/ plus focused build/test instruction files

## Notes
- No versions/flags/CI matrix values duplicated in instruction docs
- Mutable details remain in existing source-of-truth config files

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,6 +1,5 @@
 # GitHub Copilot Instructions — ScalableVectorSearch
 
-## Source of truth
 This file is the canonical instruction set for Copilot behavior in this repository
 (required flow, precedence, and fallback rules).
 


.github/instructions/general.instructions.md
mihaic, line 1: I think the general use case is covered by `.github/copilot-instructions.md` so we should remove this file.

AGENTS.md
```diff
@@ -0,0 +1,46 @@
...
+  ~ limitations under the License.
+-->
+
+# AGENTS.md — ScalableVectorSearch
+
+Entry point for agent context in this repo.
+
+## Required flow
+1. Read this file.
+2. Read the nearest `AGENTS.md` for touched files.
```
mihaic: Agents with support for AGENTS.md already have these instructions. For the [cases when Copilot doesn't look at AGENTS.md](https://docs.github.com/en/copilot/reference/custom-instructions-support), we added these instructions in its own file.
```suggestion
```

AGENTS.md
```diff
@@ -0,0 +1,46 @@
...
+## Directory map
+- `.github/AGENTS.md`
+- `benchmark/AGENTS.md`
+- `bindings/AGENTS.md`
+- `cmake/AGENTS.md`
+- `examples/AGENTS.md`
+- `include/svs/AGENTS.md`
+- `tests/AGENTS.md`
+- `tools/AGENTS.md`
+- `utils/AGENTS.md`
```
mihaic: See previous comment.
```suggestion
```

.github/instructions/general.instructions.md
napetrov: Good point — removed `.github/instructions/general.instructions.md` to avoid duplication. Copilot policy now lives only in `.github/copilot-instructions.md`.

AGENTS.md
napetrov: Agreed. I trimmed `AGENTS.md` to AGENTS-aware context only and removed duplicated Copilot-policy text. Copilot-specific rules remain in `.github/copilot-instructions.md`.

AGENTS.md
napetrov: Addressed in the latest update: `AGENTS.md` now stays minimal and non-duplicative, with Copilot behavior centralized in `.github/copilot-instructions.md`.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,24 @@
+# GitHub Copilot Instructions — ScalableVectorSearch
+
+## Source of truth
```
mihaic: This heading duplicates the "Source-of-truth files" heading below. We can leave the text in this section at the top level, without a heading.
```suggestion
```

PR #280, opened 2026-03-02
Native serialization to a stream for FlatIndex

razdoburdin:
Reopening of https://github.com/intel/ScalableVectorSearch/pull/275 for developer branch

### DIFF
--- a/include/svs/lib/stream.h
+++ b/include/svs/lib/stream.h
@@ -27,7 +27,7 @@
 #include <cstdint>
 #include <sstream>
 
-namespace {
+namespace svs::lib::detail {
 template <typename T> auto get_buffer_size(T& ss) {
     if constexpr (requires { ss.rdbuf()->view(); }) {
         return ss.rdbuf()->view().size();
@@ -35,7 +35,7 @@ template <typename T> auto get_buffer_size(T& ss) {
         return ss.str().size();
     }
 }
-} // namespace
+} // namespace svs::lib::detail
 
 namespace svs::lib {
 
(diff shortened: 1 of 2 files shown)


include/svs/core/io/native.h
```diff
@@ -402,20 +414,30 @@ template <typename T = void> class Writer {
...
     size_t dimension_;
     lib::UUID uuid_;
     std::ofstream stream_;
     size_t writes_this_vector_ = 0;
-    size_t vectors_written_ = 0;
+};
+
+template <typename T = void> class StreamWriter : public Writer<T, StreamWriter<T>> {
+  public:
+    StreamWriter(std::ostream& os)
```
rfsaliev, line 434: It seemed like the `Header` structure written by `FileWriter` at the beginning of a file has some important information including:
* magic number and uuid - for versioning.
* stored data size.
Why `StreamWriter` does not populate the same header?

General question:
How are we going to handle cases when several objects to be stored/loaded in a stream?
E.g. in case of Vamana index, we have to store/load *configuration*, *graph* and *data* (where *data* may contain 2 simple datasets for LVQ/LeanVec cases).

razdoburdin: For `FileWriter` ostream is seek able (we know that it is a fstream) , so we can insert placeholder, write data, calculate size of data being written, replace placeholder to an actual header. But for `StreamWriter` ostream may be non seek able, and we can't do the same trick with placeholder Header.
I see two options here:
1. Create temporary seek able stringstream, and use it as a buffer. But it creates a 2x memory overhead in serialization.
2. Extract all required information from metadata. In this case we don't need Header. 

I have used the fhe first approach (with stringstream) for toml::table serialization, since metadata are small, and overhead doesn't look like an acceptable trade-off in this case. 
But for the main data I try to realize the second option (without overhead). I haven't started work on `Vamana` yet, so I am not confident, if metadata contains all required information in this case.

rfsaliev: So, I would add a test for flat+LVQ/LeanVec to validate if multi-dataset cases are managed properly.

PR #282, opened 2026-03-04
Enhance heuristic pruning to handle duplicate clusters

### LINKED ISSUE #80: Corner case when duplicate datapoints become the entry point
A potential issue arises when the entry point node contains a number of duplicate vectors (vectors with the same values) that are equal to or greater than the max_degree. In this scenario, the graph traversal remains stuck at the entry point because all the nearest neighbors are these duplicates, leading to a recall drop to zero

DIlkhush00:
This PR fixes #80 . It adds a post-pruning step to both: 

- `IterativePruneStrategy` 
- `ProgressivePruneStrategy`

Approach: If a duplicate cluster is detected, the last (worst) slot in the `result` is replaced with the closest candidate from the pool that does not have the same distance.

### DIFF
--- a/include/svs/index/vamana/prune.h
+++ b/include/svs/index/vamana/prune.h
@@ -207,10 +207,15 @@ void heuristic_prune_neighbors(
                     break;
                 }
             }
-            if (!in_result) {
-                result.back() = detail::construct_as(lib::Type<I>(), candidate);
-                break;
+            assert(
+                !in_result &&
+                "Candidate with non-anchor distance should not already be in result"
+            );
+            if (in_result) {
+                continue;
             }
+            result.back() = detail::construct_as(lib::Type<I>(), candidate);
+            break;
         }
     }
 }
@@ -314,10 +319,15 @@ void heuristic_prune_neighbors(
                     break;
                 }
             }
-            if (!in_result) {
-                result.back() = detail::construct_as(lib::Type<I>(), candidate);
-                break;
+            assert(
+                !in_result &&
+                "Candidate with non-anchor distance should not already be in result"
+            );
+            if (in_result) {
+                continue;
             }
+            result.back() = detail::construct_as(lib::Type<I>(), candidate);
+            break;
         }
     }
 }
--- a/include/svs/index/vamana/prune.h
+++ b/include/svs/index/vamana/prune.h
@@ -130,6 +130,9 @@ void heuristic_prune_neighbors(
 
     auto pruned = std::vector<PruneState>(poolsize, PruneState::Available);
     float current_alpha = 1.0f;
+    float anchor_dist = 0.0f;
+    bool anchor_set = false;
+    bool all_duplicates = true;
     while (result.size() < max_result_size && !cmp(alpha, current_alpha)) {
         size_t start = 0;
         while (result.size() < max_result_size && start < poolsize) {
@@ -145,6 +148,16 @@ void heuristic_prune_neighbors(
             const auto& query = accessor(dataset, id);
             distance::maybe_fix_argument(distance_function, query);
             result.push_back(detail::construct_as(lib::Type<I>(), pool[start]));
+
+            if (all_duplicates) {
+                if (!anchor_set) {
+                    anchor_dist = pool[start].distance();
+                    anchor_set = true;
+                } else if (pool[start].distance() != anchor_dist) {
+                    all_duplicates = false;
+                }
+            }
+
             for (size_t t = start + 1; t < poolsize; ++t) {
                 if (excluded(pruned[t])) {
                     continue;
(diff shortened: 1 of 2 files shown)


include/svs/index/vamana/prune.h
```diff
@@ -171,6 +184,35 @@ void heuristic_prune_neighbors(
...
+            if (cid == current_node_id || candidate.distance() == anchor_dist) {
+                continue;
+            }
+            bool in_result = false;
+            for (const auto& r : result) {
+                if (result_id(r) == static_cast<size_t>(cid)) {
+                    in_result = true;
+                    break;
+                }
+            }
```
ibhati, line 209: I’m confused why `cid` could already be in `result` here. We just checked above that this candidate's distance differs from `anchor_dist`, so I wouldn’t expect it to be present (as all the results have same distance). Should this be an assert instead (i.e., this candidate must not already be in result)? Am I missing a scenario?

ibhati (changes requested):
Thanks @DIlkhush00 for working on this issue. The solution looks good overall. I left one comment below, and I was also wondering if we could add a test that specifically triggers this case.
Additionally, could you generate a synthetic dataset with a larger number of vectors (e.g., 1M) and measure the build-time impact? It would also be helpful to benchmark GIST-1M with and without this change to confirm that:

Datasets without duplicates show no recall or build-time regression.
In scenarios where duplicates do occur, this logic does not significantly impact build time, and recall remains consistent.

Thanks again for your contribution!

include/svs/index/vamana/prune.h
DIlkhush00: Ah, right... by logic any remaining candidate shouldn't already be in `result`. I'll change this to an assert and update both strategy accordingly.

DIlkhush00:
Hi @ibhati , as suggested, I tested the build-time impact using a synthetic 1M vector dataset (128 dimensions). The results are as follows:

 - Clean Dataset: ~543s (without fix) vs. ~545s (with fix).
 - Duplicate Dataset: ~546s (with fix).

The indexing overhead seems to be negligible like under 0.5%. Also let me know if you'd like me to share the script to generate the datasets.

ibhati:
> Hi @ibhati , as suggested, I tested the build-time impact using a synthetic 1M vector dataset (128 dimensions). The results are as follows:
> 
> * Clean Dataset: ~543s (without fix) vs. ~545s (with fix).
> * Duplicate Dataset: ~546s (with fix).
> 
> The indexing overhead seems to be negligible like under 0.5%. Also let me know if you'd like me to share the script to generate the datasets.

This looks great — thanks @DIlkhush00!
As a final step before we merge, do you have a script to generate the 1M‑vector synthetic dataset you used for testing? I’d like to run it on my end to double‑check that everything behaves as expected.
Thanks again!

DIlkhush00:
Sure, here's the script:
```py

import numpy as np
import pathlib

def save_fvecs(path, data):
    """Save a numpy array to .fvecs format."""
    data = data.astype(np.float32)
    num_vectors, dim = data.shape
    dim_column = np.full((num_vectors, 1), dim, dtype=np.int32)
    with open(path, "wb") as f:
        for i in range(num_vectors):
            dim_column[i].tofile(f)
            data[i].tofile(f)

def main():
    np.random.seed(42)
    base = pathlib.Path("data/bench_1m")
    base.mkdir(parents=True, exist_ok=True)
    
    num_vectors = 1_000_000
    dim = 128
    
    print(f"Generating {num_vectors} random vectors...")
    data = np.random.random((num_vectors, dim)).astype(np.float32)
    save_fvecs(base / "db_random.fvecs", data)
    
    print("Creating duplicate-rich version (injecting clusters)...")
    data_duplicates = data.copy()

    # inject 100 clusters of 1000 identical vectors each
    for i in range(100):
        source_vector = data[i]
        start_idx = 10000 + (i * 1000)
        data_duplicates[start_idx : start_idx + 1000] = source_vector
    save_fvecs(base / "db_duplicates.fvecs", data_duplicates)
    
    print("Generating 1000 queries...")
    queries = np.random.random((1000, dim)).astype(np.float32)
    save_fvecs(base / "queries.fvecs", queries)
    print(f"Done. Files saved to {base}")

if __name__ == "__main__":
    main()


```

ibhati (approved):

ibhati:
Thanks @DIlkhush00 for the contribution and for sharing the dataset-generation script! I was able to reproduce your benchmark on my end and the results are consistent with what you reported — build-time overhead is negligible and the fix only affects graphs in the duplicate-cluster case as intended. LGTM, will get this merged shortly.

ethanglaser:
Thanks for the contribution @DIlkhush00 !

PR #285, opened 2026-03-11
Native serialization for Vamana index

razdoburdin:
This PR introduce native serialization for `Vamana` index.

Main changes are:
1. New overload of `svs::index::vamana::auto_assemble` required for direct deserialization accepts lazy loaders and call them in a flexible order to cover legacy serilized models. 
2. `save_table` method is renamed to `metadata` to avoid confusions, as far as it doesn't save anything.
3. Some minor refactoring to streamline the logic.
4. Serialization for `MutableVamana` is also implemented to avoid compilation errors. Deserialization and related tests for `MutableVamana` are expected later.

### DIFF
--- a/include/svs/core/graph/graph.h
+++ b/include/svs/core/graph/graph.h
@@ -289,8 +289,8 @@ template <std::unsigned_integral Idx, data::MemoryDataset Data> class SimpleGrap
         return table;
     }
 
-    template <class FileName_t>
-    lib::SaveTable metadata(const FileName_t& filename, const lib::UUID& uuid) const {
+    template <class FileName>
+    lib::SaveTable metadata(const FileName& filename, const lib::UUID& uuid) const {
         auto table = metadata();
         table.insert("binary_file", filename);
         table.insert("uuid", uuid.str());


include/svs/core/graph/graph.h
```diff
@@ -276,22 +276,36 @@ template <std::unsigned_integral Idx, data::MemoryDataset Data> class SimpleGrap
...
              {"max_degree", lib::save(max_degree())},
              {"num_vertices", lib::save(n_nodes())},
-             {"uuid", lib::save(uuid.str())},
              {"eltype", lib::save(datatype_v<Idx>)}}
         );
+        return table;
+    }
+
+    template <class FileName_t>
+    lib::SaveTable metadata(const FileName_t& filename, const lib::UUID& uuid) const {
```
rfsaliev: Nitpick: would be better to follow existing template parameters naming style:


```suggestion
    template <class FileName>
    lib::SaveTable metadata(const FileName& filename, const lib::UUID& uuid) const {
```

include/svs/index/vamana/index.h
```diff
@@ -814,6 +814,20 @@ class VamanaIndex {
...
     }
 
+    void save(std::ostream& os) const {
+        // Construct and save runtime parameters.
+        auto parameters = VamanaIndexParameters{
+            entry_point_.front(), build_parameters_, get_search_parameters()};
+
+        lib::begin_serialization(os);
+        // Config
+        lib::save_to_stream(parameters.save(), os);
```
rfsaliev: Does it make sense to rename `VamanaIndexParameters::save()` to `metadata()` as well?
Or implement the infrastructure code for `save_to_stream()` similar to `save_to_disk()` then call just
`lib::save_to_stream(parameters, os)`?

include/svs/lib/saveload/load.h
```diff
@@ -967,13 +971,25 @@ T load_from_stream(
...
+        );
+    } else {
+        return lib::load(
+            loader,
+            detail::read_metadata(deserializer, stream),
+            deserializer,
+            stream,
+            SVS_FWD(args)...
+        );
+    }
```
rfsaliev: Got some doubts how this condition looks like.
Doesn't it make sense to define 2 specializations of `load_from_stream` using newly defined concept `ContextFreeLoadable` which `requires T::load(...`?
...
I would also look for possibility of modifying/extending `LoadContext` infrastructure and related classes to store the `{deserializer, stream}` information in a context.

include/svs/index/vamana/index.h
razdoburdin: done

include/svs/lib/saveload/load.h
razdoburdin: I have created 2 specializations for `load_from_stream`

The existing infrastructure for `LoadContext`/`SaveContext` doesn't take the data order into account. Organizing of the input/output stream is made by `DirectoryArchiver`. But for native streaming we want to avoid (or at least minimize) creating of temporary buffers, that means that we have to control the read/write order. This is the reason I have chosen not to reuse this logic in native streaming.

PR #287, opened 2026-03-13
feat(ivf): Add numpy array support for IVF assemble functions

ibhati:
This enables users to build IVF indices from in-memory numpy arrays without intermediate file I/O, improving performance for dynamic workflows.
- Add assemble_from_clustering and assemble_from_file methods that accept numpy arrays directly (py_data parameter) in addition to file paths
- Refactor assemble_dynamic_from_clustering to avoid double data copy:
      - Internal _impl function takes data by const reference
      - Smart dispatching detects ImmutableMemoryDataset and passes by reference
- Add timing instrumentation for assemble operations
- Add comprehensive tests for numpy array assembly in both IVF and DynamicIVF
- Zero-copy path: numpy views passed directly without data duplication

### DIFF
--- a/include/svs/index/ivf/dynamic_ivf.h
+++ b/include/svs/index/ivf/dynamic_ivf.h
@@ -1003,16 +1003,18 @@ class DynamicIVFIndex {
     }
 };
 
-/// @brief Assemble a DynamicIVFIndex from clustering and data prototype
+/// @brief Internal implementation for assembling a DynamicIVFIndex.
+///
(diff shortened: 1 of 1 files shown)


bindings/python/src/ivf.cpp
```diff
@@ -685,7 +874,6 @@ are automatically detected from the saved configuration file.
 
     /// Index building
     // Build from Numpy array.
-    detail::add_build_specialization<svs::BFloat16>(clustering);
```
ethanglaser, line 688: Why is this one removed?

PR #289, opened 2026-03-16
Native serialization for MultiMutableVamana index

razdoburdin:
This PR introduce native serialization for `MultiMutableVamana` index.
Should be merged after: https://github.com/intel/ScalableVectorSearch/pull/286

Main changes are:

1. New overload of `svs::index::vamana::auto_multi_dynamic_assemble` required for direct deserialization accepts lazy loaders and call them in a flexible order to cover legacy serialized models.
2. Added related tests.
3. `supports_saving` flag is keeped false, as far as it isn't used.
4. `MultiMutableVamana` doesn't have an orchestrator. So no changes on this side.

### DIFF
--- a/include/svs/index/vamana/multi.h
+++ b/include/svs/index/vamana/multi.h
@@ -563,16 +563,8 @@ class MultiMutableVamanaIndex {
     constexpr std::string_view name() const { return "multi dynamic vamana index"; }
 
     static constexpr lib::Version save_version = lib::Version(0, 0, 0);
-    void save(
-        const std::filesystem::path& config_directory,
-        const std::filesystem::path& graph_directory,
-        const std::filesystem::path& data_directory
-    ) {
-        // Post-consolidation, all entries should be "valid".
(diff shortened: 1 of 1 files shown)


include/svs/index/vamana/multi.h
```diff
@@ -628,6 +628,57 @@ class MultiMutableVamanaIndex {
...
+        });
+
+        size_t num_labels = ext_lab_vec.size();
+        std::vector<label_type> labels(num_labels);
+        std::transform(
+            ext_lab_vec.begin(),
+            ext_lab_vec.end(),
+            labels.begin(),
+            [](const auto& ext_lab) { return ext_lab.second; }
+        );
```
rfsaliev: Does it make sense to extract this preparation code to a separate method and reuse it both in tis method and in `save(const filesystem::path&, const filesystem::path&, const filesystem::path)`?

razdoburdin: done

PR #292, opened 2026-03-19
Separate native and legacy deserialization paths

razdoburdin:
This PR diverse deserialization paths for newly serialized models with a legacy ones. 

The main changes are:
1. Legacy models are now deserialized with intermediate files, as it was done before.
2. Native deserialization path is cleaned, as far as we don't need support of legacy models for this path.


rfsaliev (approved):
LGFM

PR #297, opened 2026-03-23
[CPP Runtime] Refactor storage tags and add static Vamana index support

rfsaliev:
## Enable static Vamana index functionality support in CPP Runtime API.

This PR includes following changes:
1. Refactored CPP Runtime storage factory utils to support configurable storage allocator.  
  * `StorageKindTag<Kind>` is removed and replaced with `StorageType<Kind, Alloc>` in usages
  * `Storagefactory::init()` `block_size` argument is replaced with `allocator` - to be constructed in index implementation code (with appropriate block size)
2. [static] `VamanaIndex` interface implementation.
3. Refactored IVF storage factory utils to match changes made in step 1.

### DIFF
--- a/bindings/cpp/src/svs_runtime_utils.h
+++ b/bindings/cpp/src/svs_runtime_utils.h
@@ -159,9 +159,12 @@ template <typename Alloc>
 Alloc make_allocator(svs::lib::PowerOfTwo blocksize_bytes)
     requires(svs::data::is_blocked_v<Alloc>)
 {
-    assert(
-        blocksize_bytes.raw() > 0 && "Blocked storage types require a non-zero blocksize"
-    );
+    if (blocksize_bytes.raw() == 0) {
+        throw StatusException(
+            ErrorCode::INVALID_ARGUMENT,
+            "Blocked storage types require a non-zero blocksize"
+        );
+    }
     auto parameters = svs::data::BlockingParameters{.blocksize_bytes = blocksize_bytes};
     return Alloc(parameters);
 }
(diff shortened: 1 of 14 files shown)


ibhati (approved):

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -123,71 +123,100 @@ inline bool is_supported_storage_kind(StorageKind kind) {
...
+};
 
 // Storage factory
 template <typename T> struct StorageFactory;
 
 // Unsupported storage type factory returning runtime error when attempted to be used.
 // Return type defined to simple to allow substitution in templates.
-template <> struct StorageFactory<UnsupportedStorageType> {
-    using StorageType = SimpleDatasetType<float>;
+template <typename Alloc> struct StorageFactory<UnsupportedStorageType<Alloc>> {
```
razdoburdin, line 211: Is possible to unify StorageFactory with IVFStorageFactory?

razdoburdin (approved):

bindings/cpp/src/svs_runtime_utils.h
rfsaliev: Yes, it is possible, but I would like to do such refactoring in a separate PR - this one is big enough already :)
@ibhati, your comments?

rfsaliev: Reported the issue #303

ibhati: Yes, it is definitely possible. But I agree, let's do it in a separate PR.

PR #300, opened 2026-03-24
Native serialization for Indexes.

razdoburdin:
This PR adds native stream serialization to all SVS index types, as an alternative to the existing (legacy)  directory-based serialization. It allow to avoid filesystem round-trips of the data. The native serialization doesn't require from the stream to be seek able, so no additional restrictions were introduced.

See the following PR for details: https://github.com/intel/ScalableVectorSearch/pull/280, https://github.com/intel/ScalableVectorSearch/pull/281, https://github.com/intel/ScalableVectorSearch/pull/285, https://github.com/intel/ScalableVectorSearch/pull/286, https://github.com/intel/ScalableVectorSearch/pull/289, https://github.com/intel/ScalableVectorSearch/pull/292, https://github.com/intel/ScalableVectorSearch/pull/294, https://github.com/intel/ScalableVectorSearch/pull/296, https://github.com/intel/ScalableVectorSearch/pull/299

Main changes are:
1. A CRTP base `Archiver` extracts binary I/O primitives (`write_size`, `read_size`, `write_name`, `read_name`, `read_from_istream`) from `DirectoryArchiver`. `DirectoryArchiver` and new `StreamArchiver` class inherit from `Archiver`. `StreamArchiver` has its own magic number ("SVS_STRM") to distinguish native streams from directory archives.
2. The monolithic `Writer` is split via CRTP with two derived classes: `FileWriter` owns an `std::ofstream`, writes a header, flushes on destructor, `StreamWriter` wraps an external `std::ostream&`, no header/lifecycle management. This allows `io::save(data, os)` to write vector data directly to any stream.
3. The `save(stream)` in orchestrator `Impl` classes no longer does temp-dir->pack. Instead it directly calls `impl().save(stream)`.
4. The dispatching between new (native) and old (legacy) deserialization is made at the orchestrators. `Deserializer::build(is)` reads the magic number, exposes `is_native()` to choose path.


rfsaliev (approved):
LGFM

PR #302, opened 2026-03-25
Add MMapAllocator for SSD-backed memory-mapped data allocation

ibhati:
Add MMapAllocator that allocates memory via mmap-backed files on a configurable path (e.g. NVMe/SSD). This enables placing secondary data structures on SSD while keeping primary data in RAM, reducing heap memory usage for large-scale vector search.

- include/svs/core/allocator_mmap.h: MMapAllocator with configurable base path, access hints (Sequential/Random/Normal), and automatic file cleanup on deallocation.
- tests/svs/core/test_allocator_mmap.cpp: Unit tests for allocation, deallocation, file creation, and access hint propagation.
- tests/CMakeLists.txt: Register mmap allocator test.


include/svs/core/allocator_mmap.h
```diff
@@ -0,0 +1,393 @@
...
+    void deallocate(void* ptr, size_t SVS_UNUSED(n)) {
+        detail::MMapAllocationManager::deallocate(ptr);
+    }
+
+    ///
+    /// @brief Construct an object
+    ///
+    /// Performs default initialization of the object.
+    ///
+    void construct(T* ptr) { ::new (static_cast<void*>(ptr)) T; }
```
rfsaliev, line 320: Seems like this code is incompatible with declared `use_existing_file()` behavior, which means "read-only".

include/svs/core/allocator_mmap.h
```diff
@@ -0,0 +1,393 @@
...
+    void use_existing_file(std::filesystem::path path, size_t offset = 0) {
+        override_file_ = std::move(path);
+        override_offset_ = offset;
+    }
+
+    ///
+    /// @brief Allocate memory
+    ///
+    /// Creates a memory-mapped file and returns a pointer to it.
+    /// If use_existing_file() was called, maps that file read-only instead.
```
rfsaliev, line 260: It seems like would be better to separate behavior for in-directory an 'existing file' allocations to different classes.

include/svs/core/allocator_mmap.h
```diff
@@ -0,0 +1,393 @@
...
+
+namespace detail {
+
+///
+/// @brief Manager for file-backed memory mapped allocations
+///
+/// Tracks memory-mapped allocations by keeping MMapPtr objects alive.
+/// Thread-safe for concurrent allocations.
+///
+class MMapAllocationManager {
```
rfsaliev, line 44: It looks like this class is fully-static (all fields are static).
And actually behaves as hidden global variables + helper functions.
According to that, there is no reason to define non-static methods like `allocate()` and `map_existing_at_offset()`.
Suggesting to make all methods static and delete .ctor
Or implement the `Singleton` pattern.

rfsaliev (commented):
Design suggestions

ibhati:
> Design suggestions

@rfsaliev I made the changes as per your suggestion. Please have another look
Here is the commit id: https://github.com/intel/ScalableVectorSearch/pull/302/changes/07b3af08055e0908fe5fb46ac1002810e2e841f0

PR #309, opened 2026-04-02
add adaptive batch size heuristic for filtered search

yuejiaointel:
Currently the filtered k-NN search loop uses batch_size = k when calling iterator.next(). When the filter is restrictive (e.g., 1% of IDs pass), this results in many expensive graph traversal rounds to collect enough valid results.

  This PR introduces a heuristic that adapts the batch size based on observed filter hit rate:
  - First round uses search_window_size as the initial batch size
  - After each round, computes: batch_size = remaining_needed / hit_rate
  - If no hits yet, keeps using the initial batch size

  For example, with k=10 and a 10% filter pass rate: instead of ~100 rounds of 10 candidates, it converges in ~2 rounds.


bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -136,6 +153,8 @@ class DynamicVamanaIndexImpl {
                             }
                         }
                     }
+                    batch_size =
+                        compute_filtered_batch_size(found, k, total_checked, batch_size);
```
rfsaliev: Good idea, but, from performance perspective, I would slightly change the code:
1. Compute the batch size at the beginning of the `do-while` loop - it will avoid computation when `found==k`
2. Increment `total_checked` out-of the `for` loop.
3. It might make sense to set initial batch size the `max` of `k` and `search_window_size`
E.g.
```suggestion
                size_t total_checked = 0;
                auto batch_size = std::max(k, sp.buffer_config_.get_search_window_size());
                do {
                    batch_size =
                        compute_filtered_batch_size(found, k, total_checked, batch_size);
                    iterator.next(batch_size);
                    for (auto& neighbor : iterator.results()) {
                        if (filter->is_member(neighbor.id())) {
                            result.set(neighbor, i, found);
                            found++;
                            if (found == k) {
                                break;
                            }
                        }
                    }
                    total_checked += iterator.size();

```

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -38,6 +38,20 @@
...
+// On subsequent rounds, estimates how many candidates are needed to find the
+// remaining results given the observed hit rate.
+inline size_t compute_filtered_batch_size(
+    size_t found, size_t needed, size_t total_checked, size_t initial_batch_size
+) {
+    if (found == 0 || found >= needed) {
+        return initial_batch_size;
+    }
+    double hit_rate = static_cast<double>(found) / total_checked;
+    return static_cast<size_t>((needed - found) / hit_rate);
```
rfsaliev: I would also try to improve performance here:
* FP64 computation is not very performant
* Computation precision is not very important here
* There is potential issues in SVS BatchIterator in case of huge batch size

So, I would use the following formula:
* `hit_rate_inv = 1 / hit_rate = checked / found`
* `result = (needed - found) / hit_rate = (needed - found) * hit_rate_inv = needed * checked / found - checked`
*  The formula `needed * checked / found - checked` is most precise, but there is the bigger risk of overflow for huge `needed` and `checked` values
```suggestion
    auto hit_rate = total_checked / found + 1; // found == 0 is handled above; +1 to increase result eliminating INT precision issues
    return (needed - found) * hit_rate % max_batch_size; // max_batch_size - constant
```

Alternative (assuming, that FP32 is fast enough):
```suggestion
    float new_batch_size = static_cast<float>(needed) * total_checked / found - total_checked;
    return static_cast<size_t>(new_batch_size) % max_batch_size;
```

rfsaliev (changes requested):
Thank you for the good proposal.

**Requested changes:**
> Please apply such improvements to `range_search()` and in `vamana_index_impl.h as well

**Suggestions:**
There are some performance related suggestions in comments.

But during the review, I found, that: `compute_filtered_batch_size()` logic is prediction of further amount of processing based on previous processing results and requested amount of matches aka:
`PredictFurtherProcessing(processed, hits, goal)`
So, I would declare this function more generic, and move it to utilities header with more common signature and reuse in vamana_index_impl.h as well:
> In such case, ` % max_batch_size` operation should be applied outside of this function
```cpp
/// @param processed - number of already processed elements (total_checked)
/// @param hits - number of matched elements (found)
/// @param goal - number of requested elements to be matched (needed)
/// @param hint - result to be returned if prediction is failed, e.g. other params == 0
size_t predict_further_processing(size_t processed, size_t hits, size_t goal, size_t hint) {
    if (processed * hits * goal == 0) {
        return hint;
    }
    // use prediction formula below
    ...
}
```

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: thx added, probably need to run some benchmarks before knowing exact performance

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: Thx, added these change

bindings/cpp/src/vamana_index_impl.h
```diff
@@ -131,8 +131,20 @@ class VamanaIndexImpl {
                 auto query = queries.get_datum(i);
                 auto iterator = get_impl()->batch_iterator(query);
                 size_t found = 0;
+                size_t total_checked = 0;
+                // Use adaptive batch sizing: start with at least k candidates,
+                // then adjust based on observed filter hit rate.
+                auto batch_size = std::max(k, sp.buffer_config_.get_search_window_size());
+                const auto max_batch_size = batch_size;
```
rfsaliev: IMHO, the `max_batch_size` value should be (compile-time?) constant based on the generic SVS Vamana performance instead of current `k` or `search_window_size`.
For example:
* `k == search_window_size == 10`
* `filter->is_member()` returns `true` for 10% of results
* after first iteration, the `predict_` function will return `(10 - 1) * 10 / 1 == 90`
* but next `batch_size` will be limited to `max_batch_size == 10`
* So, we will have 10 small iterations instead of 1 big enough

rfsaliev (commented):
It seems like `max_batch_size` calculation issue.

bindings/cpp/src/vamana_index_impl.h
yuejiaointel: thx for the suggestion and agreed, the cap is too restrictive, removed the cap entirely and added a filter_stop early exist heuristic instead : 
if hit rate falls below thresold (set by user) after getting some hits, we give up and return empty, and iterator should be able to handle large batch size by growing the search buffer

yuejiaointel: there are some discussions about this during the benchmark results discussion , pulled you in to the chat

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -118,15 +118,21 @@ class DynamicVamanaIndexImpl {
...
                 auto query = queries.get_datum(i);
                 auto iterator = impl_->batch_iterator(query);
                 size_t found = 0;
+                size_t total_checked = 0;
+                auto batch_size = std::max(k, sp.buffer_config_.get_search_window_size());
                 do {
-                    iterator.next(k);
+                    batch_size =
+                        predict_further_processing(total_checked, found, k, batch_size);
+                    iterator.next(batch_size);
```
rfsaliev, line 153: What will happen on the second iteration in case if:
* filter_stop = 0.0
* batch_size = k = 100
* found = 1
?
How big bach_size will be here?

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -136,6 +142,10 @@ class DynamicVamanaIndexImpl {
                             }
                         }
                     }
+                    if (should_stop_filtered_search(total_checked, found, filter_stop)) {
+                        found = 0;
```
rfsaliev, line 165: Does it mean that we should return no results even if some results are "found"?

rfsaliev (commented):
Some questions.

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: batch size will be (100-1)*100/1 = 9900, with 1% hit to find remaining 99 results, we need 9900 more, is that too large?

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: now the logic is discard results we already found to give a clear fallback sign, if user gets 1 out of 100 requested results, it could be confusing SVS giving up vs only 1 result existed in dataset

bindings/cpp/src/dynamic_vamana_index_impl.h
rfsaliev: @ibhati , can you please clarify if batch_size=9900 is suitable for Vamana BatchIterator?
Thank you.

bindings/cpp/src/dynamic_vamana_index_impl.h
rfsaliev: Ok

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: checked with Ishwar and was told max size should be up to the number of vectors in the index, changed, thx for this question!

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -118,15 +118,23 @@ class DynamicVamanaIndexImpl {
         // Selective search with IDSelector
         auto old_sp = impl_->get_search_parameters();
         impl_->set_search_parameters(sp);
+        const float filter_stop = params ? params->filter_stop : 0.0f;
```
rfsaliev: It seems like this construction enforce user to always provide proper `filter_stop` value in case when user want to configure `SearchParams`.
To eliminate this issue, I would recommend to initialize `filter_stop` field in `SearchParams` with `Unspecify<float>` and use `set_if_specified()` here:

```suggestion
        float filter_stop = svs_default_filter_stop_defined_somewhere;
        if (params) {
            set_if_specified(filter_stop, params->filter_stop);
        }
```

bindings/cpp/include/svs/runtime/vamana_index.h
```diff
@@ -40,6 +40,11 @@ struct VamanaSearchParameters {
     size_t search_buffer_capacity = Unspecify<size_t>();
     size_t prefetch_lookahead = Unspecify<size_t>();
     size_t prefetch_step = Unspecify<size_t>();
+    // Minimum filter hit rate to continue filtered search.
+    // If the hit rate after the first round falls below this threshold,
+    // stop and return empty results (caller can fall back to exact search).
+    // Default 0 means never give up.
+    float filter_stop = 0.0f;
```
rfsaliev:
```suggestion
    float filter_stop = Unspecify<float>();
```

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -118,15 +118,23 @@ class DynamicVamanaIndexImpl {
...
 
         auto search_closure = [&](const auto& range, uint64_t SVS_UNUSED(tid)) {
             for (auto i : range) {
                 // For every query
                 auto query = queries.get_datum(i);
                 auto iterator = impl_->batch_iterator(query);
                 size_t found = 0;
+                size_t total_checked = 0;
+                auto batch_size = std::max(k, sp.buffer_config_.get_search_window_size());
+                const auto max_batch_size = impl_->size();
```
rfsaliev: This can be moved out of the `search_closure`

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -431,6 +431,32 @@ auto dispatch_storage_kind(StorageKind kind, F&& f, Args&&... args) {
 }
 } // namespace storage
 
+// Predict how many more items need to be processed to reach the goal,
+// based on the observed hit rate so far.
+// If no hits yet, returns `hint` unchanged.
+// Result is capped at `max_batch_size` (e.g., number of vectors in the index).
+inline size_t predict_further_processing(
+    size_t processed, size_t hits, size_t goal, size_t hint, size_t max_batch_size
```
rfsaliev:
```suggestion
    size_t processed, size_t hits, size_t goal, size_t hint, size_t max_value
```

rfsaliev (commented):
More comments/suggestions

bindings/cpp/src/svs_runtime_utils.h
yuejiaointel: thx, implemented these changes

rfsaliev (approved):
LGFM

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -431,6 +431,32 @@ auto dispatch_storage_kind(StorageKind kind, F&& f, Args&&... args) {
...
+    float batch_size = static_cast<float>(goal - hits) * processed / hits;
+    return std::min(std::max(static_cast<size_t>(batch_size), size_t{1}), max_value);
+}
+
+// Check if the filtered search should stop early based on the observed hit rate.
+// Returns true if the hit rate is below the threshold, meaning the caller should
+// give up and let the caller fall back to exact search.
+inline bool
+should_stop_filtered_search(size_t total_checked, size_t found, float filter_stop) {
+    if (filter_stop <= 0 || total_checked == 0 || found == 0) {
```
mihaic: Shouldn't the search stop if `found` is 0 after a first search?

yuejiaointel: Currently we only check hit rate after finding some results, idea is it could just be bad luck in initial query

mihaic: My initial idea was to give up as early as possible; that includes giving up when nothing is found after the first try. I think `found == 0` should not be a special case.

yuejiaointel: got it fixed, now found = 0 will just give up

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -118,15 +118,26 @@ class DynamicVamanaIndexImpl {
...
+        const auto max_batch_size = impl_->size();
 
         auto search_closure = [&](const auto& range, uint64_t SVS_UNUSED(tid)) {
             for (auto i : range) {
                 // For every query
                 auto query = queries.get_datum(i);
                 auto iterator = impl_->batch_iterator(query);
                 size_t found = 0;
+                size_t total_checked = 0;
+                auto batch_size = std::max(k, sp.buffer_config_.get_search_window_size());
```
mihaic: Maybe you remember I was also proposing to estimate the initial batch size by checking the hit rate of a random sample of nodes. In case the user does not adjust the search window size based on the filter, we can adjust it automatically based on this sample hit rate. What do you think? We can make it opt-in with a boolean `filter_estimate_batch` parameter. We can also use this sample hit rate with `filter_stop` and return even before the first search.

yuejiaointel: Yeh I think it will be a great addition, added, logic is getting default 200 random ids from index to get initial hit rate

yuejiaointel: default is enable filter_estimate_batch

bindings/cpp/src/svs_runtime_utils.h
```diff
@@ -431,6 +432,113 @@ auto dispatch_storage_kind(StorageKind kind, F&& f, Args&&... args) {
...
+    size_t actual_sample = std::min(sample_size, pool_size);
+    std::mt19937 rng(42);
+    std::uniform_int_distribution<size_t> dist(0, pool_size - 1);
+    size_t hits = 0;
+    for (size_t i = 0; i < actual_sample; ++i) {
+        if (filter.is_member(id_lookup(dist(rng)))) {
+            hits++;
+        }
+    }
+    return static_cast<float>(hits) / actual_sample;
```
rfsaliev: in case when this function returns `std::make_pair(actual_sample, hits)`,
`compute_initial_batch_size()` can be simply replaced with:
 `predict_further_processing(actual_sample, hits, k, std::max(k, search_window_size), max_size)`
and `should_stop_filtered_search_by_estimate()` - with:
`should_stop_filtered_search(actual_sample, hits, filter_stop)`

rfsaliev: Such change would lead significant simplification of `search()` methods and potentially avoid the issue:

## bindings/cpp/src/vamana_index_impl.h#L91-L207
> Complex Method

yuejiaointel: thx, refactored

mihaic (approved):

PR #315, opened 2026-04-08
[CPP Runtime] Add get_distance and reconstruct_at to VamanaIndex API

yuejiaointel:
## Summary
- Add `get_distance()` and `reconstruct_at()` methods to the runtime shared library (`VamanaIndex` and `DynamicVamanaIndex`)
- `get_distance` computes distance between a stored vector (by ID) and a query vector, returning a `double`
- `reconstruct_at` decompresses/reconstructs stored vectors back to `float32` by ID — works transparently with all storage kinds (FP32, FP16, SQI8, LVQ, LeanVec)
- Both methods follow the existing runtime API pattern: pure virtual in the public header, `runtime_error_wrapper` in the manager, orchestrator dispatch in the impl layer


bindings/cpp/include/svs/runtime/vamana_index.h
```diff
@@ -76,6 +76,13 @@ struct SVS_RUNTIME_API VamanaIndex {
         IDFilter* filter = nullptr
     ) const noexcept = 0;
 
+    // Compute distance between stored vector `id` and `query` (dim floats).
+    virtual Status get_distance(double* distance, size_t id, const float* query)
```
rfsaliev: Not sure if it makes sense to add `double` type usage for just 1 call.
Wouldn't it be enough `float` type here?
@ibhati , your thoughts?

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -305,6 +305,23 @@ class DynamicVamanaIndexImpl {
...
+        auto query_span = std::span<const float>(query, dim_);
+        return impl_->get_distance(id, query_span);
+    }
+
+    void reconstruct_at(size_t n, const size_t* ids, float* output) {
+        if (!impl_) {
+            throw StatusException{ErrorCode::NOT_INITIALIZED, "Index not initialized"};
+        }
+        svs::data::SimpleDataView<float> dst{output, n, dim_};
+        std::span<const uint64_t> id_span{reinterpret_cast<const uint64_t*>(ids), n};
```
rfsaliev, line 321: Even if on x86-64 CPU `size_t` is equal to `uint64_t`, there is still the risk of `reinterpret_cast` here.
Would suggest type-safe conversion:

```suggestion
        std::vector<const uint64_t> id_vec(ids, ids + n);
```

bindings/cpp/src/dynamic_vamana_index_impl.h
```diff
@@ -305,6 +305,23 @@ class DynamicVamanaIndexImpl {
...
 
+    double get_distance(size_t id, const float* query) const {
+        if (!impl_) {
+            throw StatusException{ErrorCode::NOT_INITIALIZED, "Index not initialized"};
+        }
+        auto query_span = std::span<const float>(query, dim_);
+        return impl_->get_distance(id, query_span);
+    }
+
+    void reconstruct_at(size_t n, const size_t* ids, float* output) {
```
rfsaliev, line 316: Instead of raw pointers here, I would recommend to follow approaches for existing methods' signatures defined in this class where arguments are typed.
See: `void DynamicVamanaIndexImpl::add(data::ConstSimpleDataView<float> data, std::span<const size_t> labels)`

rfsaliev (changes requested):
Please change new methods' signatures in `VamanaIndexImpl` and `DynamicVamanaIndexImpl` according to the signature style used in other existing methods i.e. typed containers rather than raw pointers

bindings/cpp/include/svs/runtime/vamana_index.h
yuejiaointel: That's a good point, changed to float to match the rest of the apis

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: Thx, changed

bindings/cpp/src/dynamic_vamana_index_impl.h
yuejiaointel: Thx, changed

bindings/cpp/include/svs/runtime/vamana_index.h
```diff
@@ -84,6 +84,13 @@ struct SVS_RUNTIME_API VamanaIndex {
         IDFilter* filter = nullptr
     ) const noexcept = 0;
 
+    // Compute distance between stored vector `id` and `query` (dim floats).
+    virtual Status
+    get_distance(float* distance, size_t id, const float* query) const noexcept = 0;
+
+    // Reconstruct `n` vectors by ID into `output` buffer (n * dim floats).
+    virtual Status reconstruct_at(size_t n, const size_t* ids, float* output) noexcept = 0;
```
rfsaliev: nit: slight inconsistency: in  `get_distance()`, the 'output' argument is the first, when `reconstruct_at()` - last.

rfsaliev (approved):
LGFM
nitpicking comment can be ignored or addressed - up to your perfectionism :)

bindings/cpp/include/svs/runtime/vamana_index.h
yuejiaointel: thx, changed

PR #318, opened 2026-04-16
Fix: cmake version compatibility issue in C API CMakeLists.txt

### LINKED ISSUE #317: FetchContent_Declare uses DOWNLOAD_EXTRACT_TIMESTAMP which is incompatible with CMake < 3.24
@mihaic @rfsaliev  SVS c binding make file has DOWNLOAD_EXTRACT_TIMESTAMP argument in FetchContent_Declare. This was only available in CMake 3.24 and newer. Amazon Linux ships with CMake 3.22.2. So it will trigger an error
```
CMake Error at /usr/share/cmake/Modules/ExternalProject.cmake:2806 (message):
  At least one entry of URL is a path (invalid in a list)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FetchContent.cmake:1075 (message):
    CMake step for svs failed: 1
```

Environment
CMake version: 3.22.2
Compiler: GCC 11.5.0
OS: Amazon Linux 2023 (Linux x86_64)
Build command: cmake -S bindings/c -B build -DSVS_RUNTIME_ENABLE_LVQ_LEANVEC=ON

Suggested Fix
```
if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.24")
    set(_svs_fetch_timestamp DOWNLOAD_EXTRACT_TIMESTAMP TRUE)
else()
    set(_svs_fetch_timestamp "")
endif()
FetchContent_Declare(
    svs
    URL ${SVS_URL}
    ${_svs_fetch_timestamp}
)
```

rfsaliev:
Resolve cmake version compatibility issue caused by using DOWNLOAD_EXTRACT_TIMESTAMP which is introduced in v.3.24

This PR fixes #317

### DIFF
--- a/bindings/c/CMakeLists.txt
+++ b/bindings/c/CMakeLists.txt
@@ -118,7 +118,6 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
         FetchContent_Declare(
             svs
             URL ${SVS_URL}
-            DOWNLOAD_EXTRACT_TIMESTAMP TRUE
         )
         FetchContent_MakeAvailable(svs)
         list(APPEND CMAKE_PREFIX_PATH "${svs_SOURCE_DIR}")


mihaic (approved):

rfsaliev:
@klmckeig, please review

klmckeig (approved):

PR #319, opened 2026-04-17
Update version to v0.4.0

ethanglaser:
## Summary
- Version bump from v0.3.0 to v0.4.0 across all version files
- Updated NEWS.md with v0.4.0 release notes, moved v0.3.0 notes to HISTORY.md
- Updated runtime API version (MINOR 2 -> 4) in version.h to align with project version

## Release Notes
* Native serialization support for Indexes (#300)
* Clang-21 compiler support with gcc-11 headers for shared library builds
* Amazonlinux2 support for shared/static library dropped
* Consolidated to publish only minimal necessary shared library (14 -> 6) and runtime bindings (2 -> 1) release artifacts


bindings/cpp/include/svs/runtime/version.h
```diff
@@ -41,7 +41,7 @@
 
 #ifndef SVS_RUNTIME_VERSION_MINOR
 /// Minor version number - incremented for backward-compatible feature additions
-#define SVS_RUNTIME_VERSION_MINOR 2
+#define SVS_RUNTIME_VERSION_MINOR 4
```
mihaic, line 44: Can you please explain the versioning? I was expecting 3.

mihaic (changes requested):

bindings/cpp/include/svs/runtime/version.h
ethanglaser: This was missed in the last release but I believe it should be aligned with library version. Please confirm @rfsaliev

rfsaliev: Yes, it has to be adjusted.

mihaic (approved):

PR #322, opened 2026-04-28
IVF: add IDFilter support and `set_intra_query_threads` in cpp runtime

ibhati:
Brings two FAISS-required runtime features to the IVF C++ bindings:

`IDFilter`-based search to support `SearchParameters::sel` selectors in `IndexSVSIVF`.
The ability to change the intra-query thread count after `train()` / `load()` (mirrors Vamana).

### DIFF
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -123,14 +123,14 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
     else()
         # Links to LTO-enabled static library, requires GCC/G++ 11.2
         if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.3")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-lto-ivf.tar.gz"
+            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-lto-nightly-2026-05-06-1253.tar.gz"
                 CACHE STRING "URL to download SVS shared library")
         else()
             message(WARNING
                 "Pre-built LVQ/LeanVec SVS library requires GCC/G++ v.11.2 to apply LTO optimizations."
                 "Current compiler: ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION}"
             )
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-ivf.tar.gz"
+            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-nightly-2026-05-06-1253.tar.gz"
                 CACHE STRING "URL to download SVS shared library")
         endif()
         include(FetchContent)
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -121,18 +121,9 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
             "Please build SVS Runtime using bindings/cpp directory as CMake source root."
         )
     else()
-        # Links to LTO-enabled static library, requires GCC/G++ 11.2
-        if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.3")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-lto-ivf.tar.gz"
-                CACHE STRING "URL to download SVS shared library")
-        else()
-            message(WARNING
-                "Pre-built LVQ/LeanVec SVS library requires GCC/G++ v.11.2 to apply LTO optimizations."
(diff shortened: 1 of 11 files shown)


bindings/cpp/CMakeLists.txt
```diff
@@ -121,18 +121,9 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
             "Please build SVS Runtime using bindings/cpp directory as CMake source root."
         )
     else()
-        # Links to LTO-enabled static library, requires GCC/G++ 11.2
-        if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.3")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-lto-ivf.tar.gz"
-                CACHE STRING "URL to download SVS shared library")
-        else()
```
ethanglaser, line 128: I think this can be restored

ibhati: Done!

include/svs/orchestrators/ivf.h
```diff
@@ -73,6 +77,14 @@ class IVFImpl : public manager::ManagerImpl<QueryTypes, Impl, IFace> {
         return std::string{typename_impl.begin(), typename_impl.end() - 1};
     }
 
+    ///// Intra-query (cluster-level) parallelism
+    [[nodiscard]] size_t get_num_intra_query_threads() const override {
```
ethanglaser, line 81: Just curious on why [[nodiscard]] is used here and not elsewhere - I am not particularly familiar with what nodiscard does

bindings/cpp/src/dynamic_ivf_index_impl.h
```diff
@@ -145,7 +146,84 @@ class DynamicIVFIndexImpl {
...
+        // Pre-search filter sampling: estimate hit rate before cluster traversal.
+        size_t sampled = 0;
+        size_t sample_hits = 0;
+        const auto initial_batch_hint = std::max(k, size_t{1});
+        auto initial_batch_size = initial_batch_hint;
+        if (filter_estimate_batch) {
+            std::tie(sampled, sample_hits) = sample_filter_hits(
+                *filter,
+                max_batch_size,
+                [this](size_t id) { return impl_->has_id(id); },
```
ethanglaser, line 181: Looks like this is nearly identical diff to the non-dynamic except this line - why is it different here?

Side note - are these classes something that we should look into having inheritance for?

bindings/cpp/src/dynamic_ivf_index_impl.h
```diff
@@ -174,6 +252,18 @@ class DynamicIVFIndexImpl {
...
     }
 
+    void set_intra_query_threads(size_t intra_query_threads) {
+        if (intra_query_threads == 0) {
+            throw StatusException{
+                ErrorCode::INVALID_ARGUMENT, "intra_query_threads must be at least 1"};
+        }
+        intra_query_threads_ = intra_query_threads;
+        // Note: impl_ uses the value set at construction time,
+        // changing it here requires rebuilding/reloading the index
```
ethanglaser: This seems not ideal - should it call impl_->set_num_intra_query_threads(intra_query_threads) ?

bindings/cpp/tests/ivf_runtime_test.cpp
ethanglaser, line 1: Worth testing case where filter_estimate_batch=false?
And what about comparison of results for runs with different number of intra_query_threads?

ibhati: Agree, added. Thanks!

bindings/cpp/src/dynamic_ivf_index_impl.h
ibhati: Good catch! You're absolutely right - this is now fixed in commit d0e0d3c! 

The setter now properly calls `impl_->set_num_intra_query_threads(intra_query_threads)` and the getter returns `impl_->get_num_intra_query_threads()` when `impl_` is available. This makes the behavior consistent with Vamana and allows dynamic thread count updates after index construction.

Note: This assumes the v0.4.0 library will be available (the earlier revert to v0.3.0 URLs will be updated separately when the new release is ready).

bindings/cpp/src/dynamic_ivf_index_impl.h
ibhati: Great observation! The difference is intentional and reflects the underlying API differences:

- **DynamicIVF** uses `impl_->has_id(id)` - checks if an ID exists in the dynamic index
- **Static IVF** uses `impl_->contains(id)` - checks if an ID is present in the static index

Both are correct for their respective implementations. The dynamic variant has `has_id()` because it manages a set of active IDs that can be added/removed, while the static variant uses `contains()` which is a more general existence check.

Regarding inheritance - that's a good architectural suggestion! Right now these classes follow the existing pattern in the runtime bindings, but consolidating common functionality could be a nice refactoring for the future.

include/svs/orchestrators/ivf.h
ibhati: The `[[nodiscard]]` attribute is used consistently in the public interface methods in `include/svs/orchestrators/ivf.h` (lines 67, 76, 81, 89) where it's important to ensure callers don't ignore return values from getters like `get_search_parameters()`, `get_num_intra_query_threads()`, etc.

This is consistent with the existing codebase pattern where `[[nodiscard]]` is primarily used at the public API boundary.

bindings/cpp/CMakeLists.txt
ethanglaser: Hmm actually I may have been wrong about this...I figured since no changes to LVQ/LeanVec source occurred in this PR that the shared lib would not need to be updated, but looks like the changes in include/ do affect the shared lib. Try to bring back this change - though I think it can be done with just a single line change (https://github.com/intel/ScalableVectorSearch/blob/main/bindings/cpp/CMakeLists.txt#L126)

ethanglaser (approved):
LGTM, let's wait to confirm runtime lib CI is green

PR #324, opened 2026-04-29
Add inline, remove unused exception variable name, safe float computation

mnorris11:
This is a patch to get it to work internally. It would be easier if things are in SVS, so next version upgrade we don't need to apply this patch.

Reason for changes:
  1. inlining: This is from -Werror,-Winline-namespace-reopened-noninline.              
  2. Unused exception parameter: -Werror,-Wunused-exception-parameter. 
  3. Float change: acccording to AI: the shift 125 - e goes out of range for uint32_t when e is outside [102, 112]. The result is multiplied by zero so it's "dead" at runtime, but the shift itself is still UB. UBSAN catches it and 5 FP16 tests fail:   WriteAndReadIndexSVSFP16, VamanaFP16TrainSaveLoadAndAdd, WriteAndReadStaticIVFFP16, WriteAndReadIndexSVSIVFFP16,  IVFFP16TrainAndAdd. 
  4. index.h Thread clamping: per AI: "compute_centroid_distances (in common.h:675) partitions centroids across threads using StaticPartition(num_centroids). If you have more threads than centroids, some threads get an empty partition. The callback then calls batch.start() and batch.size() on those empty ranges, leading to OOB access into centroids.get_datum() and matmul_results[tid]. The fix clamps the thread  pool size to at most the number of centroids at construction time, so every thread gets at least one centroid to work on."
  5. sorted_buffer.h change: "  back() returns candidates_[size_ - 1]. When the buffer is empty (size_ == 0), that underflows to candidates_[SIZE_MAX] — an OOB read. The original code evaluated back() before checking full(), so on an empty buffer it would access garbage memory before full() returned false and short-circuited the &&. Swapping the operands ensures full() is checked first; if the buffer isn't full, back() is never called. Classic short-circuit evaluation fix."

### DIFF
--- a/bindings/cpp/include/svs/runtime/api_defs.h
+++ b/bindings/cpp/include/svs/runtime/api_defs.h
@@ -32,7 +32,7 @@
 
 namespace svs {
 namespace runtime {
-namespace v0 {
+inline namespace v0 {
 
 class OptionalBool {
     enum class Value : int8_t { Undef = -1, True = 1, False = 0 };
--- a/include/svs/lib/float16.h
+++ b/include/svs/lib/float16.h
@@ -56,10 +56,14 @@ inline uint16_t float_to_float16_untyped_slow(const float x) {
     const uint32_t b = bitcast_float_to_uint32(x) + 0x00001000;
     const uint32_t e = (b & 0x7F800000) >> 23; // exponent
     const uint32_t m = b & 0x007FFFFF;         // mantissa
+    // Compute denormalized shift safely: only valid when 101 < e < 113,
+    // which means shift amount (125 - e) is in range [13, 23].
+    // Clamp e to avoid undefined behavior when e > 125 or e < 102.
+    const uint32_t safe_e = (e > 101 && e < 113) ? e : 112;
     return (b & 0x80000000) >> 16 |
            static_cast<uint32_t>(e > 112) * ((((e - 112) << 10) & 0x7C00) | m >> 13) |
            static_cast<uint32_t>((e < 113) && (e > 101)) *
-               ((((0x007FF000 + m) >> (125 - e)) + 1) >> 1) |
+               ((((0x007FF000 + m) >> (125 - safe_e)) + 1) >> 1) |
            static_cast<uint32_t>(e > 143) *
                0x7FFF; // sign : normalized : denormalized : saturate
 }
--- a/include/svs/lib/threads/thread.h
+++ b/include/svs/lib/threads/thread.h
@@ -855,8 +855,10 @@ template <typename T = telemetry::NoTelemetry> class ThreadImpl {
         // Wait for the future to become available and rethrow the exception.
         wait_for_result();
         get_result();
-        throw ANNEXCEPTION("Expected to get an exception from a crashed thread but no "
-                           "exception was thrown!");
+        throw ANNEXCEPTION(
+            "Expected to get an exception from a crashed thread but no "
+            "exception was thrown!"
+        );
     }
 
     // Assign Work
@@ -888,7 +890,7 @@ template <typename T = telemetry::NoTelemetry> class ThreadImpl {
         // * Catch this exception and wrap its message inside a `ThreadError`.
         try {
             unsafe_assign(fn);
-        } catch (const ThreadCrashedError& err) {
+        } catch (const ThreadCrashedError&) {
             try {
                 unsafe_get_exception();
             } catch (const std::exception& inner_error) { throw ThreadError{inner_error}; }
--- a/bindings/cpp/include/svs/runtime/dynamic_ivf_index.h
+++ b/bindings/cpp/include/svs/runtime/dynamic_ivf_index.h
@@ -24,7 +24,7 @@
 
 namespace svs {
 namespace runtime {
-namespace v0 {
+inline namespace v0 {
 
 /// @brief Abstract interface for dynamic IVF indices (supports add/delete).
 struct SVS_RUNTIME_API DynamicIVFIndex : public IVFIndex {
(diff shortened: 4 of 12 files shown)


include/svs/lib/float16.h
```diff
@@ -56,10 +56,14 @@ inline uint16_t float_to_float16_untyped_slow(const float x) {
...
     const uint32_t m = b & 0x007FFFFF;         // mantissa
+    // Compute denormalized shift safely: only valid when 101 < e < 113,
+    // which means shift amount (125 - e) is in range [13, 23].
+    // Clamp e to avoid undefined behavior when e > 125 or e < 102.
+    const uint32_t safe_e = (e > 101 && e < 113) ? e : 112;
     return (b & 0x80000000) >> 16 |
            static_cast<uint32_t>(e > 112) * ((((e - 112) << 10) & 0x7C00) | m >> 13) |
            static_cast<uint32_t>((e < 113) && (e > 101)) *
-               ((((0x007FF000 + m) >> (125 - e)) + 1) >> 1) |
+               ((((0x007FF000 + m) >> (125 - safe_e)) + 1) >> 1) |
```
rfsaliev, line 66: Hm, in case when `(e < 101 || e > 113)`, this component is always `0`:
```cpp
return (b & 0x80000000) >> 16 |
           ... |
           static_cast<uint32_t>((e < 113) && (e > 101)) *
               ((((0x007FF000 + m) >> (125 - safe_e)) + 1) >> 1) |

// equal to:
return (b & 0x80000000) >> 16 |
           ... |
           static_cast<uint32_t>(e > 101 && e < 113) *
               ((((0x007FF000 + m) >> (125 - safe_e)) + 1) >> 1) |
// equal to:
return (b & 0x80000000) >> 16 |
           ... |
           static_cast<uint32_t>(false) *
               ((((0x007FF000 + m) >> (125 - safe_e)) + 1) >> 1) |
//equal to:
return (b & 0x80000000) >> 16 |
           ... |
           0 *
               ((((0x007FF000 + m) >> (125 - safe_e)) + 1) >> 1) |

```

include/svs/lib/threads/thread.h
```diff
@@ -888,7 +890,7 @@ template <typename T = telemetry::NoTelemetry> class ThreadImpl {
         // * Catch this exception and wrap its message inside a `ThreadError`.
         try {
             unsafe_assign(fn);
-        } catch (const ThreadCrashedError& err) {
+        } catch (const ThreadCrashedError&) {
```
rfsaliev, line 893:
```suggestion
        } catch (const ThreadCrashedError& SVS_UNUSED(err)) {
```

include/svs/lib/threads/thread.h
```diff
@@ -855,8 +855,10 @@ template <typename T = telemetry::NoTelemetry> class ThreadImpl {
         // Wait for the future to become available and rethrow the exception.
         wait_for_result();
         get_result();
-        throw ANNEXCEPTION("Expected to get an exception from a crashed thread but no "
-                           "exception was thrown!");
+        throw ANNEXCEPTION(
+            "Expected to get an exception from a crashed thread but no "
+            "exception was thrown!"
+        );
```
rfsaliev, line 861: Please, use SVS formatting rules.

bindings/cpp/include/svs/runtime/api_defs.h
```diff
@@ -32,7 +32,7 @@
 
 namespace svs {
 namespace runtime {
-namespace v0 {
+inline namespace v0 {
```
rfsaliev, line 35: IMHO, this change will negatively affect usability and maintainability for further API major versions.
Instead, a macro can be used here which value defined in `version.h` depending on `SVS_RUNTIME_API_VERSION`, e.g.:
```
#if (SVS_RUNTIME_API_VERSION == 0)
...
#define SVS_API_V0 inline namespace v0
...
#endif
```
With usage aka:
```suggestion
namespace svs {
namespace runtime {
SVS_API_V0 {
...
} // SVS_API_V0
```

rfsaliev (commented):
Thank you @mnorris11 , for the contribution.
There are some suggestions, proposals and comments

include/svs/lib/float16.h
ibhati: @rfsaliev please read this comment above:
Float change: acccording to AI: the shift 125 - e goes out of range for uint32_t when e is outside [102, 112]. The result is multiplied by zero so it's "dead" at runtime, but the shift itself is still UB. UBSAN catches it and 5 FP16 tests fail: WriteAndReadIndexSVSFP16, VamanaFP16TrainSaveLoadAndAdd, WriteAndReadStaticIVFFP16, WriteAndReadIndexSVSIVFFP16, IVFFP16TrainAndAdd.

rfsaliev: Oh, yes. My fault :)

rfsaliev:
Thank you @mnorris11 for the PR.
Unfortunately, more complicated changes had to be applied for proper fixes.
There are created 3 separate PRs based on your proposal:
* #328
* #329
* #330

mnorris11:
> Thank you @mnorris11 for the PR. Unfortunately, more complicated changes had to be applied for proper fixes. There are created 3 separate PRs based on your proposal:
> 
> * [[CPPRuntime] Fix C++ Runtime Clang compilation errors #328](https://github.com/intel/ScalableVectorSearch/pull/328)
> * [Refactor thread pool creation to validate thread count against centroids #329](https://github.com/intel/ScalableVectorSearch/pull/329)
> * [Refactor float_to_float16_untyped_slow for clarity and optimization #330](https://github.com/intel/ScalableVectorSearch/pull/330)

gotcha, let me know if I should do anything on this PR!

ethanglaser:
@mnorris11 are you using clang compiler internally? Or can you provide some more details about what circumstances led to these failures? We can try to align our CI with these cases to avoid issues moving forward.

rfsaliev:
Closing, because implementation made in #328, #329, #330 have been merged to main.
Thank you @mnorris11 for the contribution.

PR #326, opened 2026-05-05
feat(vamana): deferred compression for dynamic Vamana index

ibhati:
## Summary
Adds support for "deferred compression" on `svs::DynamicVamana`: the index can be built initially with an uncompressed backend (FP32/FP16) and transparently swap to a compressed target backend (SQ8 / LVQ / LeanVec) once a configured live-vector threshold is crossed. The already-built graph, ID translator, status, and entry point are reused — only the underlying `Data` is retrained and transplanted.

## Motivation
Compressed backends (LVQ/LeanVec) require sufficient training data for good recall. With purely-eager dynamic builds, users either (a) build with poor compression on a tiny initial population or (b) rebuild from scratch later. Deferred compression resolves this: build cheaply on an uncompressed backend, then swap once enough vectors are present.

## Public API additions
- **`svs::index::vamana::MutableVamanaIndex::TransplantTag`** — new constructor + tag that builds a `MutableVamanaIndex` over a *new* `Data` while reusing an existing graph / status / entry-point / translator. Validates `data.size() == graph.n_nodes() == status.size()`.
- **Rvalue-qualified `release_data()` / `release_graph()` / `release_status()` / `release_entry_point()` / `release_translator()`** on `MutableVamanaIndex` — move-out accessors enabling the transplant.
- **`svs::DynamicVamana::get_typed_impl<QueryTypes, Impl>()`** — controlled type-erasure escape hatch returning the concrete `MutableVamanaIndex<...>*` (or `nullptr` on type mismatch).
- **`svs::runtime::v0::VamanaIndex::DynamicIndexParams`**:
  - `size_t deferred_compression_threshold = 0` (0 disables; >0 arms).
  - `StorageKind initial_storage_kind = StorageKind::FP32` (must be FP32 or FP16).
- **`svs::runtime::v0::DynamicVamanaIndex::get_current_storage_kind()`** — observable transition between initial and target kind.

## Runtime behavior
- Default (`threshold == 0`): unchanged eager behavior.
- Threshold > 0: builds the initial backend with `initial_storage_kind`, installs a per-source-tag swap closure that retrains and transplants when `size() >= threshold`.
- **Fast path**: if the very first `add()` already meets the threshold, the runtime skips the staging build and constructs the target compressed backend directly.

## Out of scope
- Python bindings (separate follow-up).
- Persistence of deferred config (saved indices remain post-swap state).

### DIFF
--- a/include/svs/index/vamana/dynamic_index.h
+++ b/include/svs/index/vamana/dynamic_index.h
@@ -300,52 +300,69 @@ class MutableVamanaIndex {
     /// See `MutableVamanaIndex(TransplantTag, ...)` for the rationale.
     struct TransplantTag {};
 
-    /// @brief Transplant constructor: build a `MutableVamanaIndex` over a freshly
-    /// constructed `Data` while reusing graph/translator/status/entry_point from a prior
-    /// instance.
-    ///
-    /// This is the building block for "deferred compression": once enough vectors have
-    /// been accumulated in an uncompressed dynamic index, the caller can train a
-    /// compressed dataset from the accumulated vectors, then transplant the existing
-    /// graph (and bookkeeping structures) onto a new index whose `Data` template
-    /// parameter is the compressed type, avoiding a graph rebuild.
-    ///
-    /// Preconditions
-    /// * ``data.size() == graph.n_nodes() == status.size()``
-    /// * ``translator`` covers exactly the slots marked ``SlotMetadata::Valid`` (or
-    ///   ``Deleted``) in ``status``.
-    /// * The graph type and distance type are compatible with the new ``data``.
-    ///
-    /// The new instance recomputes ``first_empty_`` from ``status``.
-    template <typename Pool>
+  private:
+    /// @brief Internal aggregate of the per-instance state moved out of the source
+    /// index during a transplant. Constructed by `_take_transplant_inputs` and
+    /// consumed by the private delegating constructor below. Not part of the
+    /// public API.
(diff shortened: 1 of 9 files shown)


include/svs/index/vamana/dynamic_index.h
```diff
@@ -295,6 +295,79 @@ class MutableVamanaIndex {
...
+    ///
+    /// Preconditions
+    /// * ``data.size() == graph.n_nodes() == status.size()``
+    /// * ``translator`` covers exactly the slots marked ``SlotMetadata::Valid`` (or
+    ///   ``Deleted``) in ``status``.
+    /// * The graph type and distance type are compatible with the new ``data``.
+    ///
+    /// The new instance recomputes ``first_empty_`` from ``status``.
+    template <typename Pool>
+    MutableVamanaIndex(
```
rfsaliev, line 345: I would think about kind of 'move' .ctor with signature like:
```cpp
template <typename SourceData, typename Pool>
MutableVamanaIndex(
  MutableVamanaIndex<Graph, SourceData, Dist>&& source_index,
  Pool threadpool,
 ...
)
 graph_{source_index.release_graph()},
 ...
{
...
}
```

In such case, we do not need user code to know about `MutableVamanaIndex` internal structure and which data to be moved.

ibhati: Thanks I will make this change

eshenayo:
@rfsaliev is this change going to merge?

PR #338, opened 2026-05-22
Add primary-only LeanVec storage kinds to C++ runtime

ibhati:
> Supersedes #323 (which couldn't be reopened via API after force-push).

## Summary

Adds the **public C++ runtime surface** for primary-only LeanVec. This is the public-side counterpart to the private re-architecture; see the private PR for the design discussion and rationale.

## Changes

- `StorageKind` enum: two new variants `LeanVecLVQ4PrimaryOnly` and `LeanVecLVQ8PrimaryOnly`.
- `is_leanvec_storage()`: updated to recognize them.
- `LeanDatasetType` alias + `StorageType` specializations: produce `LeanDataset<UsingLVQ<N>, void, ...>` for the new kinds (mapping to the private partial specialization).
- `StorageFactory` specializations: build/load the primary-only LeanDataset (no secondary tier).
- `SVS_DISPATCH_STORAGE_KIND` macro: add both new cases to the dispatch.

## Architectural alignment

The private side now uses a partial specialization `LeanDataset<T1, void, ...>` (no runtime flag, no secondary member, no secondary methods) — mirroring how LVQ has separate one-level vs two-level classes. The public side exposes this through new `StorageKind` enum values, which is the same way LVQ4/LVQ8 variants are already exposed.

## Diff

+100 / −1 across 5 files. All additive at the public surface: no existing enum values changed, no existing API signatures changed.

### DIFF
--- a/bindings/cpp/include/svs/runtime/api_defs.h
+++ b/bindings/cpp/include/svs/runtime/api_defs.h
@@ -106,6 +106,10 @@ enum class StorageKind {
     LeanVec4x4,
     LeanVec4x8,
     LeanVec8x8,
+    // Primary-only LeanVec variants: only the lossy primary tier is stored, no
+    // secondary (reranking) tier. Trades reranking accuracy for ~50% less memory.
+    LeanVecLVQ4PrimaryOnly,
+    LeanVecLVQ8PrimaryOnly,
 };
 
 enum class ErrorCode {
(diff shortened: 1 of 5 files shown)


rfsaliev (approved):
LGFM

> Just the question: aren't enum value names aka `LeanVecLVQ4PrimaryOnly` too long?

PR #345, opened 2026-06-29
[Vamana] Add get_memory_usage() to report allocated bytes

yuejiaointel:
## Summary

Adds `get_memory_usage()` returning the total number of bytes a Vamana index has **allocated** — graph storage + vector data + metadata — so an integrator can accurately report and bound SVS memory consumption.

Both the static `VamanaIndex` and the dynamic `MutableVamanaIndex` are covered, and the method is plumbed through the orchestrator layers (`VamanaInterface` virtual → `VamanaImpl` override → `Vamana` / `DynamicVamana`) so it is callable on `svs::Vamana` and `svs::DynamicVamana`.

## Why

Integrators (e.g. memory-bounded module hosts) need to account for memory SVS allocates via `mmap`/blocked allocators, which bypass the host's `malloc` accounting. The existing `blocksize_bytes()` reports only the initial block size and is neither an upper nor lower bound on the real footprint. `get_memory_usage()` reports the true allocated total across all blocks plus metadata.

## What

Accounting is **capacity-based** (the bytes the containers have reserved, not just live elements) so that block over-allocation is reflected:

| Component | Source |
|---|---|
| `graph_bytes` | `graph_.get_data().capacity() * element_size()` |
| `data_bytes` | `data_.capacity() * data_.element_size()` |
| `metadata_bytes` (dynamic only) | slot-status vector + entry-point list + estimate of the external/internal ID translation maps |

A `VamanaMemoryUsage { graph_bytes, data_bytes, metadata_bytes, total() }` struct and `get_memory_breakdown()` expose the per-component split; `get_memory_usage()` returns `get_memory_breakdown().total()`.

Notes:
- A `detail::dataset_allocated_bytes()` helper uses capacity-based accounting when the dataset exposes `capacity()` (flat/blocked `SimpleData`), and falls back to live element count otherwise (e.g. `SQDataset`). No public accessors or signatures were changed.
- The ID-translation map byte size is not directly queryable; it is estimated from the entry count (accurate to within a few percent), with a comment noting the approximation.

## Tests

New unit tests at both the core-index and orchestrator levels for the static and dynamic indices:
- `tests/svs/index/vamana/index.cpp`, `tests/svs/index/vamana/dynamic_index.cpp`
- `tests/svs/orchestrators/vamana.cpp`, `tests/svs/orchestrators/dynamic_vamana.cpp`

Assertions: usage `> 0` for a built index, breakdown components sum to the total, and `graph_bytes`/`data_bytes` are non-zero. `[managers]` and the touched index tags pass with no regressions.

### DIFF
--- a/bindings/cpp/src/dynamic_vamana_index.cpp
+++ b/bindings/cpp/src/dynamic_vamana_index.cpp
@@ -215,8 +215,9 @@ Status DynamicVamanaIndex::build(
     *index = nullptr;
 
     auto status = DynamicVamanaIndex::check_params(dynamic_index_params);
-    if (!status.ok())
+    if (!status.ok()) {
         return status;
+    }
 
     return runtime_error_wrapper([&] {
         auto impl = std::make_unique<Impl>(
@@ -306,8 +307,9 @@ Status DynamicVamanaIndexLeanVec::build(
     *index = nullptr;
 
     auto status = DynamicVamanaIndex::check_params(dynamic_index_params);
-    if (!status.ok())
+    if (!status.ok()) {
         return status;
+    }
 
     return runtime_error_wrapper([&] {
         auto impl = std::make_unique<Impl>(
@@ -338,8 +340,9 @@ Status DynamicVamanaIndexLeanVec::build(
     *index = nullptr;
 
     auto status = DynamicVamanaIndex::check_params(dynamic_index_params);
-    if (!status.ok())
+    if (!status.ok()) {
         return status;
+    }
 
     return runtime_error_wrapper([&] {
         auto training_data_impl =
--- a/bindings/cpp/src/dynamic_vamana_index.cpp
+++ b/bindings/cpp/src/dynamic_vamana_index.cpp
@@ -171,11 +171,13 @@ Status DynamicVamanaIndex::check_params(
     constexpr static size_t kMaxBlockSizeExp = 30; // 1GB
     constexpr static size_t kMinBlockSizeExp = 12; // 4KB
 
-    if (dynamic_index_params.blocksize_exp > kMaxBlockSizeExp)
+    if (dynamic_index_params.blocksize_exp > kMaxBlockSizeExp) {
         return Status(ErrorCode::INVALID_ARGUMENT, "Blocksize is too large");
+    }
 
-    if (dynamic_index_params.blocksize_exp < kMinBlockSizeExp)
+    if (dynamic_index_params.blocksize_exp < kMinBlockSizeExp) {
         return Status(ErrorCode::INVALID_ARGUMENT, "Blocksize is too small");
+    }
 
     return Status_Ok;
 }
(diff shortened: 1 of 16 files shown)


yuejiaointel:
current failing for runtime test, it should be gone after this change goes into nightly

include/svs/index/vamana/index.h
```diff
@@ -177,6 +177,32 @@ struct VamanaIndexParameters {
...
 
+namespace detail {
+
+/// @brief Return the number of bytes allocated for the backing storage of ``dataset``.
+///
+/// Capacity-based accounting (the bytes the containers have reserved) is used whenever the
+/// dataset exposes a ``capacity()`` accessor (e.g. flat and blocked ``SimpleData``), so
+/// that block over-allocation is reflected. Datasets that do not expose ``capacity()``
+/// fall back to the number of live elements.
+template <typename Dataset> size_t dataset_allocated_bytes(const Dataset& dataset) {
```
rfsaliev: IMHO, it makes sense moving this helper to a common header, e.g. `svs/core/data.h`

bindings/cpp/src/dynamic_vamana_index.cpp
```diff
@@ -65,6 +65,17 @@ struct DynamicVamanaIndexManagerBase : public DynamicVamanaIndex {
 
     size_t blocksize_bytes() const noexcept override { return impl_->blocksize_bytes(); }
 
+    size_t get_memory_usage() const noexcept override { return impl_->get_memory_usage(); }
```
rfsaliev, line 68: the statement `impl_->get_memory_usage()` should be wrapped to `runtime_error_wrapper`.
P.S. existing `blocksize_bytes()` has the same issue.

include/svs/index/vamana/index.h
```diff
@@ -744,6 +770,24 @@ class VamanaIndex {
...
+        MemoryBreakdown usage{};
+        usage.graph_bytes = detail::dataset_allocated_bytes(graph_.get_data());
+        usage.data_bytes = detail::dataset_allocated_bytes(data_);
+        usage.metadata_bytes =
+            entry_point_.capacity() * sizeof(typename entry_point_type::value_type);
+        return usage;
+    }
+
+    /// @brief Return the total number of bytes allocated by this index.
+    size_t get_memory_usage() const { return get_memory_breakdown().total(); }
```
rfsaliev: As far as the all needed information can be got from `get_memory_breackdown()`, it does not look like such method should be defined at this level.
IMHO, `get_memory_usage()` can be avoided at all or implemented at CPP Runtime level

include/svs/index/vamana/index.h
yuejiaointel: Done, moved to `svs/core/data.h`.

bindings/cpp/src/dynamic_vamana_index.cpp
yuejiaointel: `get_memory_usage()` returns `size_t`, not `Status`, so it can't be wrapped (same as `blocksize_bytes()`). The wrapped entry point is now `get_memory_breakdown()`; `get_memory_usage()` is just its total.

include/svs/index/vamana/index.h
yuejiaointel: Agreed. Removed it from core/orchestrator; `get_memory_breakdown()` is the source of truth there. Kept the total only in the runtime bindings, computed as `get_memory_breakdown().total()`.

rfsaliev (approved):
LGFM

PR #347, opened 2026-07-03
spinlock.h: add powerpc* support

pkubaj:
This is the only change needed for powerpc* support. Tested on powerpc64le.


pkubaj:
The MacOS failure doesn't seem to be related to this PR. This PR only adds code under #ifdef __powerpc__ guard and it's not likely that __powerpc__ is defined on MacOS on arm64.

PR #351, opened 2026-07-13
Refactor Blocked class to meet allocator requirements

rfsaliev:
This pull request refactors the `Blocked` class template to inherit from its allocator type instead of storing it as a member, and updates the associated test to use a struct-based allocator. This change simplifies allocator handling and improves compatibility with standard allocator patterns.

**Core class refactoring:**

* `Blocked` now inherits from its allocator type (`Alloc`) instead of storing an `Alloc allocator_` member, which simplifies construction, copying, and access to the allocator. The `get_allocator()` method now returns `*this` (as an allocator), and constructors have been updated accordingly.

**Test improvements:**

* The test for `Blocked` with an allocator has been updated to use a struct-based allocator (`I`), which provides a `value_type` and integer value for testing propagation and compatibility with the new inheritance-based implementation.

### DIFF
--- a/include/svs/core/data/simple.h
+++ b/include/svs/core/data/simple.h
@@ -648,31 +648,32 @@ struct BlockingParameters {
     std::optional<lib::PowerOfTwo> blocksize_elements = std::nullopt;
 };
 
-template <typename Alloc> class Blocked {
+template <typename Alloc> class Blocked : public Alloc {
   public:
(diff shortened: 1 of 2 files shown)


yuejiaointel (approved):
LGTM, understanding is using class instead of field saves memory because field costs at least 1 byte and class can be empty.

Issue #333, opened 2026-05-19
[C API] Add memory estimation functions for Vamana index build and search


klmckeig:
## Context

We need C API functions to estimate memory requirements before building a Vamana index. Currently we have no way to gate the build on available memory — we can only fail after the fact, after the user has already waited through a full table scan.

## Use Case

Our goal is a pre-flight check in `vamanabuild.c`:

1. Estimate peak memory required given vector count, dimensions, and index parameters
2. Check whether `maintenance_work_mem` (or system memory) is sufficient
3. Build if yes; abort early with a clear error if no

## Requested API

```c
/**
 * Estimate peak memory (bytes) needed to build a Vamana index.
 *
 * @param builder       Configured builder handle (algorithm + storage set)
 * @param num_vectors   Number of vectors to be indexed
 * @param dimensions    Vector dimensionality
 * @param out_err       Error output handle
 * @return              Estimated peak memory in bytes, or 0 on error
 */
SVS_API size_t svs_index_estimate_build_memory(
    svs_index_builder_h builder,
    size_t num_vectors,
    size_t dimensions,
    svs_error_h out_err
);

/**
 * Estimate steady-state memory (bytes) of the built index (for search).
 *
 * @param builder       Configured builder handle (algorithm + storage set)
 * @param num_vectors   Number of vectors to be indexed
 * @param dimensions    Vector dimensionality
 * @param out_err       Error output handle
 * @return              Estimated in-memory index size in bytes, or 0 on error
 */
SVS_API size_t svs_index_estimate_search_memory(
    svs_index_builder_h builder,
    size_t num_vectors,
    size_t dimensions,
    svs_error_h out_err
);
```
Taking `svs_index_builder_h` as input means the estimates automatically reflect the configured storage type (float32/16, LVQ, LeanVec, SQ) and algorithm parameters (graph degree, window size, etc.) without us duplicating that logic in the extension.

## Memory Components

**Build-time peak** (`svs_index_estimate_build_memory`):
- Graph adjacency list: `num_vectors × (graph_max_degree + 1) × sizeof(Idx)`
- Raw vector buffer: `num_vectors × dimensions × sizeof(float)` (needed before quantization)
- Temporary candidate pools / search history (scales with `window_size`, `use_full_search_history`, thread count)

**Steady-state / search** (`svs_index_estimate_search_memory`):
- Graph adjacency list (same formula)
- Stored vectors after compression (depends on storage type)
[...]

PR #354, opened 2026-07-16
[C API] Add memory accounting (get_memory_usage / get_memory_breakdown)

yuejiaointel:
## Summary

Exposes memory accounting in the **C API** for the Valkey-search integration:

- `svs_index_get_memory_usage(index, size_t* out_bytes, err)` — total allocated bytes.
- `svs_index_get_memory_breakdown(index, svs_memory_breakdown_t* out, err)` — `{graph_bytes, data_bytes, metadata_bytes}` component split.
~~- `svs_index_element_size(index, size_t* out_bytes, err)` — bytes per stored vector.~~ (keep at data level)

All follow the existing C API conventions (out-param + `svs_error_h`, `wrap_exceptions`), matching the Phase-A design in the memory-accounting contract (intel-innersource #333).

## Layers

- **C API** (`bindings/c`): the three functions + `svs_memory_breakdown_t` in `svs_c.h`; interface virtuals + concrete overrides in `src/index.hpp`; impls in `src/svs_c.cpp`.
- **Core / orchestrator**: brings in `get_memory_breakdown()` (`MemoryBreakdown` struct + capacity-based `svs::data::detail::dataset_allocated_bytes` helper) on `VamanaIndex` / `MutableVamanaIndex` and through the orchestrator, plus an `element_size()` accessor parallel to `dimensions()`. This mirrors the approved public PR #345 so the C API can build and test standalone; once #345 lands on `dev/c-api`, this reduces to just the C API layer.

## Tests

`bindings/c/tests/c_api_index.cpp` (static) and `c_api_dynamic_index.cpp` (dynamic): usage > 0, breakdown total == usage, `graph_bytes`/`data_bytes` > 0 (metadata > 0 for dynamic), `element_size == sizeof(float) * dimensions`, and null-arg handling. Both test cases pass (84 / 166 assertions).

Related: builds on intel/ScalableVectorSearch#345; memory-accounting contract in intel-innersource #333 / #326.

### DIFF
--- a/include/svs/index/vamana/dynamic_index.h
+++ b/include/svs/index/vamana/dynamic_index.h
@@ -449,13 +449,6 @@ class MutableVamanaIndex {
     ///
     size_t dimensions() const { return data_.dimensions(); }
 
-    /// @brief Return the element size (bytes per vector) for the indexed data.
-    ///
-    /// Returns the per-vector memory footprint = dataset row + graph adjacency row.
-    size_t element_size() const {
-        return data_.element_size() + graph_.get_data().element_size();
-    }
-
     /// @brief Return memory breakdown for the index.
     ///
     /// Reports the allocated memory for graph, data, and metadata components. Uses
--- a/include/svs/index/vamana/dynamic_index.h
+++ b/include/svs/index/vamana/dynamic_index.h
@@ -449,6 +449,35 @@ class MutableVamanaIndex {
     ///
     size_t dimensions() const { return data_.dimensions(); }
 
+    /// @brief Return the element size (bytes per vector) for the indexed data.
+    size_t element_size() const { return data_.element_size(); }
+
+    /// @brief Return memory breakdown for the index.
+    ///
+    /// Reports the allocated memory for graph, data, and metadata components. Uses
+    /// capacity-based accounting for datasets that expose ``capacity()``, so that block
+    /// over-allocation is reflected. Metadata includes status array, entry points, and an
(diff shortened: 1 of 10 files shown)


include/svs/index/vamana/dynamic_index.h
```diff
@@ -449,6 +449,35 @@ class MutableVamanaIndex {
     ///
     size_t dimensions() const { return data_.dimensions(); }
 
+    /// @brief Return the element size (bytes per vector) for the indexed data.
+    size_t element_size() const { return data_.element_size(); }
```
rfsaliev: IMHO, `<Index>::element_size()` should return at least `data_.element_size() + graph_.element_size()`
@eshenayo, can you please add your comment? Thank you.

P.S. It would be good to account `metadata_.element_size()` - but I am not sure if it is possible due to metadata complexity (e.g. id_translator is the map with the complex internal structure)

rfsaliev (dismissed):
It seems like `element_size()` to be reviewed

include/svs/index/vamana/dynamic_index.h
yuejiaointel: Metadata isn't per-vector (id-translator is a shared map), so it can't go in `element_size()`. And data + graph + metadata is exactly what `get_memory_breakdown()` already returns. Proposal: keep `element_size()` at the dataset level, use `get_memory_breakdown()` on the index.

rfsaliev (approved):
LGFM

PR #357, opened 2026-07-21
Update SVS_URL to nightly with get_memory_breakdown

yuejiaointel:
Bumps the pinned LTO prebuilt SVS library (`bindings/cpp/CMakeLists.txt`) from `svs-shared-library-lto-nightly-2026-05-21-1429` to `svs-shared-library-lto-nightly-2026-07-21-127`, which includes `get_memory_usage()` / `get_memory_breakdown()` (#345).

## Why

The `Build and unit tests for C++ runtime bindings (with static library, ON)` job links the runtime bindings against this prebuilt lib. The previously-pinned nightly predated #345, so it failed:

```
bindings/cpp/src/dynamic_vamana_index_impl.h:73: error: no member named 'get_memory_breakdown' in 'svs::DynamicVamana'
```

The new nightly was built from `main` (post-#345, via the private-repo submodule bump intel-innersource#338) and ships the method — verified the tarball's headers contain `get_memory_breakdown`. This should turn that CI job green.

Follows the pattern of #311 (Update SVS_URL in binaries).

PR #358, opened 2026-07-28
[C API] Expose leanvec ood

ethanglaser:
## Summary

Exposes LeanVec out-of-distribution (OOD) training to the C API. Callers can train dimensionality-reduction matrices from a data sample plus a sample of queries, then build an index that uses those matrices instead of computing PCA matrices at build time.

## API

```c
// Train matrices. Pass num_queries=0 / x_q=NULL for in-distribution (PCA).
svs_leanvec_training_data_h svs_leanvec_training_data_build(
    svs_index_builder_h builder, size_t leanvec_dims,
    size_t num_vectors, const float* x,
    size_t num_queries, const float* x_q,
    svs_error_h out_err);

void svs_leanvec_training_data_free(svs_leanvec_training_data_h);

// LeanVec storage using pre-trained matrices; leanvec_dims comes from the
// training data. Storage holds its own reference, so the handle may be freed
// as soon as this returns.
svs_storage_h svs_storage_create_leanvec_trained(
    svs_leanvec_training_data_h training_data,
    svs_data_type_t primary, svs_data_type_t secondary,
    svs_error_h out_err);
```

Training data is attached at storage construction rather than via an index-builder setter (per review): `leanvec_dims` has a single source of truth, storage configs stay immutable and safely shareable across builders, and misuse on a non-LeanVec
storage is a compile-time error instead of a runtime one.

## Tests

`c_api_index.cpp`: OOD build + search, in-distribution pre-trained build + search (freeing the training data before build, to exercise shared ownership of the matrices), and NULL-training-data rejection. All sections no-op gracefully on builds or hardware without LeanVec support.

## Notes

- Renamed internal `StorageLeanVec::lenavec_dims` → `leanvec_dims`.
- LeanVec paths are compile- and ABI-verified locally but not runtime-verified: the C bindings' pinned `SVS_URL` predates `MemoryBreakdown` (the C++ bindings were bumped in #357, the C ones weren't), and the only nightly that has it is LTO-built and won't link against local GCC 11.4. Relying on CI with private sources here.


bindings/c/include/svs/c_api/svs_c.h
```diff
@@ -333,6 +359,20 @@ SVS_API bool svs_index_builder_set_storage(
...
 
+/// @brief Attach trained LeanVec matrices to the index builder
+/// @param builder The index builder handle
+/// @param training_data The trained LeanVec matrices to use when reducing the data.
+/// Only applies when the builder's storage is configured for LeanVec; the reduced
+/// dataset is built using these matrices instead of computing PCA matrices at build
+/// time. Pass NULL to clear a previously attached training data.
+/// @param out_err An optional error handle to capture errors
+/// @return true on success, false on failure
+SVS_API bool svs_index_builder_set_leanvec_training_data(
```
rfsaliev: It seems like the "training data" belongs to LeanVec storage kind only.
The proper API call should be connected to LeanVec storage only:
Preffered:
```c
SVS_API svs_storage_h svs_storage_create_leanvec_pre_trained(
    svs_leanvec_training_data_h training_data, // leanvec_dims to be extracted from there
    svs_data_type_t primary,
    svs_data_type_t secondary,
    svs_error_h out_err /*=NULL*/
);
```
alternative:
```c
SVS_API bool svs_storage_set_leanvec_training_data(
    svs_storage_h leanvec_storage,
    svs_leanvec_training_data_h training_data, // leanvec_dims to be extracted from there and compared with previously defined
    svs_error_h out_err /*=NULL*/
);
```

rfsaliev (changes requested):
API concerns

bindings/c/include/svs/c_api/svs_c.h
ethanglaser: Updated, but named it `svs_storage_create_leanvec_trained`

rfsaliev:
@ethanglaser, it seems this PR contains number of changes out-of `bindings/c` directory.
To avoid potential conflict with the main branch, I would postpone this PR till `dev/c-api` be merged to the `main` branch.

rfsaliev (approved):
LGFM

PR #360, opened 2026-08-06
configure C API CI aligned with runtime bindings

ethanglaser:


rfsaliev (commented):
Looks good except:
`error` handle can be reused, so constructions like:
```
svs_error_free(error);
error = svs_error_create();
```
should be removed

rfsaliev (approved):
LGreatFM

PR #364, opened 2026-08-13
[C API] Implement memory estimation functions for index build and search

### LINKED ISSUE #333: [C API] Add memory estimation functions for Vamana index build and search

rfsaliev:
This pull request introduces a set of enhancements to the C API and internal implementation to support memory usage estimation for index building and search operations. The primary focus is on adding new estimation functions, implementing the underlying logic for various data builders, and integrating these features with both static and dynamic Vamana index types. The changes also include updates to the build system to incorporate new source files.

**New C API for Memory Estimation:**

- Added several new API functions to estimate memory usage for index building and search operations, including support for both static and dynamic indices, as well as search queries. These functions allow users to obtain detailed memory breakdowns before performing expensive operations.

**Implementation of Estimation Logic:**

- Implemented `estimate_size` methods for all major data builders (`SimpleDataBuilder`, `LVQDataBuilder`, `LeanVecDataBuilder`, `SQDataBuilder`) to provide accurate memory usage estimation based on builder configuration, number of vectors, and dimension. [[1]](diffhunk://#diff-f3ba11abdd57c387a19589945930e35c4fd7593343cd74ce1538771f07228d87R62-R70) [[2]](diffhunk://#diff-71f42890738a05fa27b2b395a74c294ed8f00801ed65f35b6e43ca973b53da29R83-R132) [[3]](diffhunk://#diff-811943a906bc2b6d50496a4ccb2d9e6ecda932c1b5aba09246bf0e1e0dd36dfeR79-R117) [[4]](diffhunk://#diff-610950847add2fb9ebb89679db6f18c3f1d1faa946d98a17bc0e49a972e953d7R60-R74)
- Created `data_builder.cpp` and corresponding header to dispatch and specialize estimation logic for different data types and block sizes. [[1]](diffhunk://#diff-f7358a5659b4e66947d1692af0c6b7cf11c7fecaacbb0fa1907bb5988ed51ce7R1-R133) [[2]](diffhunk://#diff-2a806c2239ebfd77eed3295a2743eb29dc6a0cfcc2fbf20869b308e5fef09f8aR22-R29)

**Integration with Vamana Index Types:**

- Added dispatcher functions to estimate memory usage for both static and dynamic Vamana indices, providing structured breakdowns of graph, data, and metadata memory requirements. [[1]](diffhunk://#diff-4d0f273362e0b22bbf59bbd72315c7e68b9205b002495c4d063412be95cd1555R132-R154) [[2]](diffhunk://#diff-01f5ea0310e9f6043dab966b4c54aa16248bd69ca55e155a22ef800b45211aa9R47-R54) [[3]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R168-R213) [[4]](diffhunk://#diff-037b3945d8fbb9964a2256d45e1ec3b2e30094386fced7aa1b0a5221a0ebbe37R52-R60)

**Build System and Dependency Updates:**

- Updated `CMakeLists.txt` to include the new `data_builder.cpp` source file, ensuring the new logic is compiled and linked.
- Added necessary includes and dependencies in implementation files to support the new estimation features. [[1]](diffhunk://#diff-6fb5101eb6123f74dd77171edbe4a686fa1fd9f963c9d450f100538ec07f6ac8R21) [[2]](diffhunk://#diff-6fb5101eb6123f74dd77171edbe4a686fa1fd9f963c9d450f100538ec07f6ac8R33-R36) [[3]](diffhunk://#diff-811943a906bc2b6d50496a4ccb2d9e6ecda932c1b5aba09246bf0e1e0dd36dfeR22)

**Summary of Most Important Changes:**

**1. API Additions**
- Added new C API functions for estimating memory usage of index building and search operations, including dynamic and static variants.

**2. Memory Estimation Logic**
- Implemented `estimate_size` methods for all major data builders to enable accurate memory usage estimation. [[1]](diffhunk://#diff-f3ba11abdd57c387a19589945930e35c4fd7593343cd74ce1538771f07228d87R62-R70) [[2]](diffhunk://#diff-71f42890738a05fa27b2b395a74c294ed8f00801ed65f35b6e43ca973b53da29R83-R132) [[3]](diffhunk://#diff-811943a906bc2b6d50496a4ccb2d9e6ecda932c1b5aba09246bf0e1e0dd36dfeR79-R117) [[4]](diffhunk://#diff-610950847add2fb9ebb89679db6f18c3f1d1faa946d98a17bc0e49a972e953d7R60-R74)
- Created `data_builder.cpp` and header to centralize and dispatch estimation logic for various data types/blocking strategies. [[1]](diffhunk://#diff-f7358a5659b4e66947d1692af0c6b7cf11c7fecaacbb0fa1907bb5988ed51ce7R1-R133) [[2]](diffhunk://#diff-2a806c2239ebfd77eed3295a2743eb29dc6a0cfcc2fbf20869b308e5fef09f8aR22-R29)

**3. Vamana Index Integration**
- Added dispatcher functions for memory estimation of static and dynamic Vamana indices, including detailed breakdowns. [[1]](diffhunk://#diff-4d0f273362e0b22bbf59bbd72315c7e68b9205b002495c4d063412be95cd1555R132-R154) [[2]](diffhunk://#diff-01f5ea0310e9f6043dab966b4c54aa16248bd69ca55e155a22ef800b45211aa9R47-R54) [[3]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R168-R213) [[4]](diffhunk://#diff-037b3945d8fbb9964a2256d45e1ec3b2e30094386fced7aa1b0a5221a0ebbe37R52-R60)

**4. Build and Dependency Updates**
- Included new source files in the build system and updated includes for proper compilation and linkage. [[1]](diffhunk://#diff-70926ef33771bc40ae0882c09583f280fc1595667ce4ef5bdbfb3e94b5f2e83bR48) [[2]](diffhunk://#diff-6fb5101eb6123f74dd77171edbe4a686fa1fd9f963c9d450f100538ec07f6ac8R21) [[3]](diffhunk://#diff-6fb5101eb6123f74dd77171edbe4a686fa1fd9f963c9d450f100538ec07f6ac8R33-R36) [[4]](diffhunk://#diff-811943a906bc2b6d50496a4ccb2d9e6ecda932c1b5aba09246bf0e1e0dd36dfeR22)

### DIFF
--- a/bindings/c/CMakeLists.txt
+++ b/bindings/c/CMakeLists.txt
@@ -141,7 +141,7 @@ if (SVS_RUNTIME_ENABLE_LVQ_LEANVEC)
     else()
         # Links to LTO-enabled static library, requires GCC/G++ 11.2
         if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.3")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-lto-nightly-2026-07-21-127.tar.gz"
+            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-lto-nightly-2026-09-09-1508.tar.gz"
                 CACHE STRING "URL to download SVS shared library")
         else()
             # The fallback is correct but slower, so nothing downstream fails and CI
--- a/bindings/c/CMakeLists.txt
+++ b/bindings/c/CMakeLists.txt
@@ -47,6 +47,7 @@ set(SVS_C_API_SOURCES
     src/dispatcher_vamana.cpp
     src/dispatcher_dynamic_vamana.cpp
     src/leanvec_training_data.cpp
+    src/data_builder.cpp
 )
 
 add_library(${TARGET_NAME} SHARED
--- a/bindings/c/src/data_builder/lvq.hpp
+++ b/bindings/c/src/data_builder/lvq.hpp
@@ -110,15 +110,19 @@ class LVQDataBuilder {
             svs::c_runtime::adjust_blocked_size(num_vectors, primary_element_sz, allocator);
 
         const auto residual_element_sz = residual_element_size(dimension);
-        const auto residual_size = svs::c_runtime::adjust_blocked_size(
-            num_vectors, residual_element_sz, allocator
-        );
+        const auto residual_size = residual_element_sz > 0
+                                       ? svs::c_runtime::adjust_blocked_size(
+                                             num_vectors, residual_element_sz, allocator
+                                         )
+                                       : 0;
 
         // Assuming a single centroid for estimation purposes
         const size_t num_centroids = 1; // Assuming 1 centroid for estimation
-        // TODO: Fix the actual memory breakdown reported by index by implementing
-        // dataset_allocated_bytes() specialization for LVQDataset.
-        const size_t centroid_size = 0; // Skipping centroids for estimation
+
+        // Note: the following size is not included in the current estimate as it is
+        // not included in memory breakdown calculations in the current implementation. It
+        // can be added if needed.
+        const size_t centroid_size = 0;
         // const auto centroid_size =
         //     sizeof(typename data_type::centroid_type::element_type) * dimension;
 
--- a/bindings/c/src/data_builder/lvq.hpp
+++ b/bindings/c/src/data_builder/lvq.hpp
@@ -73,6 +73,48 @@ class LVQDataBuilder {
     load(const std::filesystem::path& path, const allocator_type& allocator = {}) {
         return svs::lib::load_from_disk<data_type>(path, allocator);
     }
+
+    static constexpr size_t primary_element_size(size_t dimension, size_t alignment = 0) {
+        using primary_type = typename data_type::primary_type;
+        using layout_type = typename primary_type::helper_type;
+        using layout_dims_type = svs::lib::MaybeStatic<data_type::extent>;
+        const auto layout_dims = layout_dims_type{dimension};
+        return primary_type::compute_data_dimensions(layout_type{layout_dims}, alignment);
+    }
+
+    static constexpr size_t residual_element_size(size_t dims) {
+        if constexpr (ResidualBits == 0) {
+            return 0;
+        } else {
+            using residual_type = typename data_type::residual_type;
+            using dims_type = svs::lib::MaybeStatic<data_type::extent>;
+            auto residual_dims = dims_type{dims};
+            return residual_type::total_bytes(residual_dims);
+        }
+    }
+
+    size_t estimate_size(
+        size_t num_vectors, size_t dimension, const allocator_type& allocator = {}
+    ) const {
+        const size_t alignment = 0; // Assuming no specific alignment for estimation
+
+        const auto primary_element_sz = primary_element_size(dimension, alignment);
+        const auto primary_size =
+            svs::c_runtime::adjust_blocked_size(num_vectors, primary_element_sz, allocator);
+
+        const auto residual_element_sz = residual_element_size(dimension);
+        const auto residual_size = svs::c_runtime::adjust_blocked_size(
+            num_vectors, residual_element_sz, allocator
+        );
+
+        const size_t num_centroids = 1; // Assuming 1 centroid for estimation
+        const auto centroid_size =
+            sizeof(typename data_type::centroid_type::element_type) * dimension;
+
+        const auto total_size =
+            primary_size + residual_size + num_centroids * centroid_size;
+        return total_size;
+    }
 };
 
 template <size_t PrimaryBits, size_t ResidualBits, typename Alloc>
--- a/bindings/c/src/data_builder/lvq.hpp
+++ b/bindings/c/src/data_builder/lvq.hpp
@@ -52,11 +52,18 @@ class LVQDataBuilder {
   public:
     LVQDataBuilder() {}
 
+    // Follow the logic of svs::leanvec::detail::PickContainer which looks like:
+    // "Use Turbo-encoding for 4-bit LVQ."
+    using Sequential = svs::quantization::lvq::Sequential;
+    using Turbo16x8 = svs::quantization::lvq::Turbo<16, 8>;
+    template <size_t Primary, size_t Residual>
+    using AutoStrategy = std::conditional_t<(Primary == 4), Turbo16x8, Sequential>;
+
     using data_type = svs::quantization::lvq::LVQDataset<
         PrimaryBits,
         ResidualBits,
         svs::Dynamic,
-        svs::quantization::lvq::Sequential,
+        AutoStrategy<PrimaryBits, ResidualBits>,
         Allocator>;
     using allocator_type = Allocator;
 
@@ -107,9 +114,13 @@ class LVQDataBuilder {
             num_vectors, residual_element_sz, allocator
         );
 
+        // Assuming a single centroid for estimation purposes
         const size_t num_centroids = 1; // Assuming 1 centroid for estimation
-        const auto centroid_size =
-            sizeof(typename data_type::centroid_type::element_type) * dimension;
+        // TODO: Fix the actual memory breakdown reported by index by implementing
+        // dataset_allocated_bytes() specialization for LVQDataset.
+        const size_t centroid_size = 0; // Skipping centroids for estimation
+        // const auto centroid_size =
+        //     sizeof(typename data_type::centroid_type::element_type) * dimension;
 
         const auto total_size =
             primary_size + residual_size + num_centroids * centroid_size;
--- a/bindings/c/src/data_builder/lvq.hpp
+++ b/bindings/c/src/data_builder/lvq.hpp
@@ -44,27 +44,27 @@
 
 namespace svs {
 
+namespace detail {
+// Follow the logic of svs::leanvec::detail::PickContainer which looks like:
+// "Use Turbo-encoding for 4-bit LVQ."
+template <size_t Primary, size_t Residual>
+using AutoLVQStrategy = std::conditional_t<
+    (Primary == 4),
+    svs::quantization::lvq::Turbo<16, 8>,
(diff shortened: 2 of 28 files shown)


klmckeig (approved):
Looks good from integration stand of point.

bindings/c/src/data_builder/lvq.hpp
```diff
@@ -52,11 +52,18 @@ class LVQDataBuilder {
...
+    using Turbo16x8 = svs::quantization::lvq::Turbo<16, 8>;
+    template <size_t Primary, size_t Residual>
+    using AutoStrategy = std::conditional_t<(Primary == 4), Turbo16x8, Sequential>;
+
     using data_type = svs::quantization::lvq::LVQDataset<
         PrimaryBits,
         ResidualBits,
         svs::Dynamic,
-        svs::quantization::lvq::Sequential,
+        AutoStrategy<PrimaryBits, ResidualBits>,
```
yuejiaointel: This affects `build()` too, not just the estimate. A quick local test shows memory goes up for 4-bit LVQ at dims that aren't a multiple of 128 — 105 → 137 B/vector at dim 192 (+30%); 768 and 1536 are unchanged.

```cpp
using namespace svs::quantization::lvq;
static_assert(Sequential::compute_bytes(4, 192) == 96);
static_assert(Turbo<16, 8>::compute_bytes(4, 192) == 128);  // +32 B/vector
static_assert(Sequential::compute_bytes(4, 768) == Turbo<16, 8>::compute_bytes(4, 768));
```

Maybe we could use a template parameter instead, put last so existing `LVQDataBuilder<P, R, Alloc>` uses still compile?

```cpp
template <
    size_t PrimaryBits,
    size_t ResidualBits,
    typename Allocator = svs::lib::Allocator<std::byte>,
    typename Strategy = svs::quantization::lvq::Sequential>
class LVQDataBuilder {
  public:
    using data_type = svs::quantization::lvq::LVQDataset<
        PrimaryBits,
        ResidualBits,
        svs::Dynamic,
        Strategy,
        Allocator>;
```

rfsaliev: Using `Sequential` strategy for all cases by default conflicts with actual strategy selection in `svs::leanvec::detail::PickContainer`.
Prefer to follow existing strategy selection in this wrapper for LVQ dataset rather than rewrite the LVQ implementation which can negatively affect performance.

bindings/c/CMakeLists.txt
ethanglaser:
```suggestion
            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-lto-nightly-2026-09-09-1508.tar.gz"
```

yuejiaointel (approved):
lgtm!

include/svs/index/vamana/dynamic_search_buffer.h
```diff
@@ -114,6 +117,38 @@ template <typename Idx, typename Cmp = std::less<>> class MutableBuffer {
     explicit MutableBuffer(size_t size, Cmp compare = Cmp{}, bool enable_visited = false)
         : MutableBuffer{SearchBufferConfig{size}, std::move(compare), enable_visited} {}
 
+    /// Estimate heap memory footprint of a search buffer with the given configuration.
+    static constexpr size_t estimate_memory_footprint(
+        const SearchBufferConfig& config, bool enable_visited = false
+    ) {
+        // The SearchBuffer contains a vector of candidates and an optional visited set. The
+        // size of the vector is determined by the total capacity of the buffer.
+        const auto candidates_num = config.get_total_capacity();
```
ethanglaser, line 126: This assumes `candidates_` never exceeds `get_total_capacity()`, which doesn't hold for `MutableBuffer`. `candidates_` stores valid *and* deleted neighbors, but `valid_capacity_` (and `full()`, via `valid() == target_capacity()`) counts only the valid ones — so with soft-deleted entries in the graph the vector grows past the `candidates_{valid_capacity_}` reservation and reallocates.

Concretely, in `unsafe_insert` the trimming branches are all gated on `slack() >= 0`, i.e. `valid() >= valid_capacity_`. Until the buffer collects `valid_capacity_` valid neighbors, nothing trims, and `insert_inner` grows `candidates_` by one per accepted candidate. Peak size is roughly `valid_capacity_ / (1 - deleted_fraction)`, and with geometric vector growth the actual allocation can be several times this estimate.

So `svs_index_builder_estimate_search_memory_dynamic` under-reports for any dynamic index with deletions — the wrong direction for a memory-budgeting API. Either scale by an expected-deleted fraction, or document that the estimate assumes no deleted entries.

bindings/c/src/index_builder.hpp
```diff
@@ -152,5 +160,155 @@ struct IndexBuilder {
...
+            size_t candidates = num_neighbors;
+            const double rate = id_filter->filter_rate();
+            if (rate > 0.0) {
+                const double needed = static_cast<double>(num_neighbors) / rate;
+                // A very low filter rate blows up the candidate buffer, which becomes the
+                // Vamana search window and makes search() prohibitively slow. Reject such
+                // configurations instead of returning a huge, unrepresentative estimate.
+                // The cap is a heuristic: search windows beyond ~1M nodes are impractical.
+                constexpr size_t MAX_FILTERED_CANDIDATES = 1'000'000;
+                if (needed > static_cast<double>(MAX_FILTERED_CANDIDATES)) {
```
ethanglaser, line 241: This rejects configurations that the real filtered search handles fine. `estimate_initial_batch_size` and `estimate_batch_size` in `filtered_search.hpp` both clamp their result to `limit` (the index size), so the search window never actually reaches `num_neighbors / filter_rate`.

For example `num_neighbors=10`, `filter_rate=1e-6` on a 1M-vector index: `needed` is 1e7, so the estimate throws, while `svs_index_search_filtered` clamps to 1M and runs. An estimation API erroring out on a workable configuration is surprising — clamping to the same `limit` the search uses (and documenting it) would match the runtime behavior.

bindings/c/src/index_builder.hpp
```diff
@@ -152,5 +160,155 @@ struct IndexBuilder {
...
+        const size_t per_buffer_size =
+            SearchBufferType::estimate_memory_footprint(
+                svs::index::vamana::SearchBufferConfig{buffer_size},
+                params.search_buffer_visited_set_
+            ) +
+            filter_overhead;
+
+        const auto threads_num = pool_builder.get_threads_num();
+        assert(threads_num > 0 && "Thread pool must have at least one thread");
+        const size_t buffers_num = std::min(threads_num, num_queries);
```
ethanglaser, line 276: `get_threads_num()` calls the user's `ops.size(self)` for `SVS_THREADPOOL_KIND_CUSTOM`, so a callback returning 0 makes `buffers_num` 0 and this function reports a 0-byte search-memory estimate. The `assert` above is compiled out in release builds.

Note `ThreadPoolBuilder::resize()` already rejects 0, and `default_threads_num()` is `max(1, hardware_concurrency())`, so the custom path is the only unguarded one. Raising `SVS_ERROR_INVALID_ARGUMENT` here (or clamping `get_threads_num()` to at least 1) would close it.

bindings/c/src/index_builder.hpp
rfsaliev: `filter_rate=1e-6` is practically too low - it means that user expects that almost 1M vectors to be filtered out to get just 1 neighbor.
And yes, processing 1M nodes in 1M Vamana graph is kind of out of common sense.
In such case, a flat array can be used.
If user's filter rate is so much low then it seems like they do something wrong and the error tell them about.

include/svs/index/vamana/dynamic_search_buffer.h
rfsaliev: Yes, dynamic index search memory underestimated in case of deletions.
I've updated the C API function documentation accordinly.

bindings/c/src/index_builder.hpp
rfsaliev: Fixed by custom thread pool validation in `threadpool.hpp`

PR #370, opened 2026-08-21
minor revisions to examples/test scope

ethanglaser:
follow-ups from samples->examples and move that CI scope from unit to bindings script


rfsaliev (approved):
LGFM

PR #372, opened 2026-08-24
refactor(dispatch): Generate and validate the dispatch surface from a single declaration

ahuber21:
The set of distance kernels compiled ahead of time is declared once, in
`cmake/dispatch-surface.cmake`, and the generated header, the per-`-march` translation
units and the ctest checks all derive from it. Configure-time validation rejects a
malformed declaration, and VNNI becomes its own ISA level so that each level's `-march`
budget matches what it promises.

Gotchas:

- **This PR now contains #373 and #375.** They were stacked on it and the review cleanup
  touched the same files, so redoing that work here rather than folding them in would have
  meant resolving the same conflicts twice. Both are closed as merged here. The diff is
  large for that reason, not because the change grew.
- **The checkers moved.** `cmake/dispatch-checks/` is now
  `tests/multi-arch/dispatch-checks/`. They are test drivers; only generation stays under
  `cmake/`.
- **These checks are `add_test`, not Catch2.** `link_probe.cpp` exists to *fail to link*
  when a declared kernel is uninstantiated, and `entry_probe.cpp` gets driven under gdb —
  neither works from inside the single `tests` binary. Porting the four `cmake -P` checkers
  to Catch2 is a follow-up.
- **`include/svs/core/distance/dispatch_surface.h` is generated but committed**, so a bare
  `-I include` compile works without cmake. CI fails if it goes stale, and asserts that a
  build with an overridden surface leaves it alone.
- **Configure now rejects two ISA levels sharing one `-march`.** Previously accepted, and it
  shipped a fault: the weaker level got compiled with instructions its runtime predicate does
  not guarantee, and the instruction checker missed it because it looks the budget up *by*
  `-march`. Not a review point; found while probing whether these checks can still fail.
- Three points from the review were declined — Python for the checkers, dropping the
  declaration checker as redundant, and parallel lists for the ISA table. Reasons are in
  the review thread.


include/svs/core/distance/dispatch_surface.h
```diff
@@ -0,0 +1,72 @@
+/*
```
rfsaliev, line 1: As I understand, this file is autogenerated from `dispatch_surface.h.in`.
Why do we need to track it in repository?
Suggesting to remove the file from repository but generate it in binary directory during config/build.

tests/multi-arch/dispatch-checks/README.md
```diff
@@ -0,0 +1,66 @@
+<!--
```
rfsaliev, line 1: Seems like files in this directory intended for tests.
IMHO it makes sense to move them to `/test`

cmake/generate-dispatch-surface.cmake
```diff
@@ -0,0 +1,250 @@
+# Copyright 2026 Intel Corporation
```
rfsaliev, line 1: Seems like the cmake code here is pretty complicated.
Writing it in form of straightforward script leads high maintenance costs.
Please, modify the code to make it more structured, e.g. split to functions.

rfsaliev (dismissed):
The change is pretty big with high review/maintenance cost.
It seems like the huge AI generated change to be reviewed by AI.
Scripts should be simplified and well structured.

tests/multi-arch/dispatch-checks/check-dispatch-linkage.cmake
```diff
@@ -0,0 +1,163 @@
...
+foreach(required SVS_PROBE_OBJECT SVS_ARCHIVE)
+    if(NOT EXISTS "${${required}}")
+        message(FATAL_ERROR "${required} does not exist: ${${required}}")
+    endif()
+endforeach()
+
+# Distance kernels, and nothing else in the archive. Mangled names are used
+# throughout: demangled ones carry `[clone .isra.0]` suffixes that differ between
+# a local instantiation and an explicit one.
+set(svs_kernel_regex "_ZN3svs8distance.*Impl")
```
rfsaliev, line 40: Mangling rules are compiler-specific.
Are you sure that the pattern will work for all supported compilers/versions?

cmake/dispatch-checks/check-dispatch-instructions.cmake
```diff
@@ -0,0 +1,148 @@
...
+
+# "<-march>|<required classes>|<forbidden classes>", one row per instruction budget
+# any ISA level declares. A budget with no row here is a hard error: an
+# unrecognized -march silently permitting everything is how a level ends up
+# emitting instructions its runtime predicate does not guarantee.
+set(svs_budget_table
+    "x86-64||ymm zmm mask vnni"
+    "haswell|ymm|zmm mask vnni"
+    "skylake-avx512|zmm|vnni"
+    "cascadelake|zmm|"
```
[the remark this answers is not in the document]
ahuber21: Will be fixed in 375 -- and all PRs will be merged together as a stack.

include/svs/core/distance/dispatch_surface.h
ahuber21: I added a default version for reference. If someone researched the code on GH only, or on a fresh checkout without prior compilation there are no missing files.

tests/multi-arch/dispatch-checks/check-dispatch-linkage.cmake
ahuber21: It's a concern I raised with the agent. Don't have the exact reply, but it was along the lines of:

- This regex will work for everything we currently use
- Checking our own emitted mangled symbols gives us more visibility/control over what we release. A change here is probably not intentional.

The second point sounded quite reasonable to me. So I figured it's worth giving this a shot. If we find the maintenance cost is too high, we can improve the tests.
From my understand though, these symbols should have been stable for the last few months and I don't see any changes on the horizon.

cmake/generate-dispatch-surface.cmake
```diff
@@ -0,0 +1,187 @@
...
+
+set(SVS_DEFAULT_DISPATCH_SURFACE_FILE "${CMAKE_CURRENT_LIST_DIR}/dispatch-surface.cmake")
+set(SVS_DISPATCH_SURFACE_FILE "${SVS_DEFAULT_DISPATCH_SURFACE_FILE}"
+    CACHE FILEPATH
+    "Declaration of the ahead-of-time distance-kernel dispatch surface"
+)
+
+# Reads the declaration and rejects it if it is malformed. Also runnable on its
+# own -- see .github/scripts/check_dispatch_surface.sh.
+set(SVS_X86_SRC_DIR "${PROJECT_SOURCE_DIR}/include/svs/multi-arch/x86")
```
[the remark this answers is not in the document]
ahuber21: `PROJECT_SOURCE_DIR` is the last `project()` in scope, not the top-level project —
that's `CMAKE_SOURCE_DIR` — and this file is only reached after `project(svs)` at
`CMakeLists.txt:19`, so it resolves to the SVS root even under `add_subdirectory()`.
Leaving as is.

Also, already widely used in the project.

cmake/multi-arch.cmake
```diff
@@ -12,25 +12,29 @@
...
+foreach(tu_spec IN LISTS SVS_DISPATCH_TU_SPECS)
+    svs_parse_tu_spec("${tu_spec}" src level arch infix)
+
+    # Carries the instruction budget for this level, and nothing else.
+    set(lib_name "svs_x86_${infix}")
     add_library(${lib_name} INTERFACE)
     target_compile_options(${lib_name} INTERFACE -march=${arch} -mtune=${arch})
-    set(obj_name ${arch}_obj)
 
+    # Named after the level, not the -march: more than one level can share an
```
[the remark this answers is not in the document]
ahuber21: Correct — the comment predated `ed22dc13`, which made duplicate `-march` values a
hard error. Fixed by rewriting the comment, since the check is the deliberate side.

cmake/generate-dispatch-surface.cmake
```diff
@@ -0,0 +1,187 @@
...
+string(APPEND SVS_GEN_LEVEL_LOOP "    /* end */")
+string(STRIP "${SVS_GEN_LEVEL_DEFINES}" SVS_GEN_LEVEL_DEFINES)
+
+#####
+##### Emit the header
+#####
+
+# The build always compiles against the build-tree copy, and it is placed ahead
+# of the source include directory so that it wins.
+set(SVS_GENERATED_INCLUDE_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated/include")
```
rfsaliev, line 104: I am still afraid that there is possible include file collision between `include/svs/core/distance/dispatch_surface.h` and `<build>/generated/include/svs/core/distance/dispatch_surface.h`

ahuber21: Would you remove the header instead or do something else entirely?

ahuber21: The intention is to generate the new header, and overwrite the existing `include/svs/core/distance/dispatch_surface.h`.
This header will be copied to `<build>` and used ... there should not be the possibility for confusion. Both headers should have identical content.

cmake/generate-dispatch-surface.cmake
```diff
@@ -0,0 +1,187 @@
...
+#####   include/svs/core/distance/dispatch_surface.h  (source tree, committed)
+#####   SVS_DISPATCH_TU_SPECS -- "<src>|<level>|<arch>|<infix>", one per ISA level
+#####
+
+include_guard(GLOBAL)
+
+include("${CMAKE_CURRENT_LIST_DIR}/dispatch-levels.cmake")
+
+set(SVS_DEFAULT_DISPATCH_SURFACE_FILE "${CMAKE_CURRENT_LIST_DIR}/dispatch-surface.cmake")
+set(SVS_DISPATCH_SURFACE_FILE "${SVS_DEFAULT_DISPATCH_SURFACE_FILE}"
```
rfsaliev, line 28: What is the motivation for this variable?

cmake/generate-dispatch-surface.cmake
```diff
@@ -0,0 +1,187 @@
...
+target_include_directories(
+    ${SVS_LIB} BEFORE INTERFACE $<BUILD_INTERFACE:${SVS_GENERATED_INCLUDE_DIR}>
+)
+
+# Refresh the committed copy too, but only for the default declaration: the
+# committed header exists so that a bare `-I include` compile works without
+# CMake, and a one-off build with an overridden surface must not rewrite it.
+# configure_file only touches the file when the content changes, so this neither
+# dirties the tree nor forces rebuilds.
+if(svs_surface_is_default)
```
rfsaliev, line 122: Seems like by default, the `<root>/include/svs/core/distance/dispatch_surface.h` is always overriden.
What is the reason having 2 identical autogenerated headers in different locations?

include/svs/multi-arch/x86/vnni.cpp
```diff
@@ -0,0 +1,32 @@
+/*
```
rfsaliev, line 1: Why this file is named `vnni.cpp`? Is it intended for avx-vnni or avx512-vnni support?

include/svs/multi-arch/x86/avx2.cpp
```diff
@@ -21,36 +21,11 @@
...
-DISTANCE_CS_INSTANTIATE_TEMPLATE(128, AVX_AVAILABILITY::AVX2);
-DISTANCE_CS_INSTANTIATE_TEMPLATE(160, AVX_AVAILABILITY::AVX2);
-DISTANCE_CS_INSTANTIATE_TEMPLATE(200, AVX_AVAILABILITY::AVX2);
-DISTANCE_CS_INSTANTIATE_TEMPLATE(512, AVX_AVAILABILITY::AVX2);
-DISTANCE_CS_INSTANTIATE_TEMPLATE(768, AVX_AVAILABILITY::AVX2);
-DISTANCE_CS_INSTANTIATE_TEMPLATE(Dynamic, AVX_AVAILABILITY::AVX2);
+// Define every kernel for this ISA level at every generated extent. Extents come
+// from cmake/dispatch-surface.cmake, type pairs from multi-arch/x86/preprocessor.h.
+#define SVS_DEFINE_FOR_DIM(DIM) SVS_INSTANTIATE_DISTANCES(template, DIM, AVX2)
+SVS_FOR_EACH_SUPPORTED_DIM(SVS_DEFINE_FOR_DIM)
```
rfsaliev, line 27: `SVS_INSTANTIATE_DISTANCES` macro is defined in `preprocessor.h`
`SVS_FOR_EACH_SUPPORTED_DIM` - in `dispatch_surface.h`
but these headers are not included.

Seems like it breaks at least readability "... If someone researched the code on GH only,..."

rfsaliev (commented):
Still some questions.

cmake/generate-dispatch-surface.cmake
ahuber21: It's there so the validator can be pointed at a fixture instead of the real declaration — that's how .github/scripts/check_dispatch_surface.sh runs tests in tests/cmake/dispatch-surface/. Nothing in a normal build sets it. Added a comment saying so.

cmake/generate-dispatch-surface.cmake
ahuber21: The header will be removed

cmake/generate-dispatch-surface.cmake
ahuber21: Header removed. So IDE will now show an error for include distance_core.h until cmake ran

include/svs/multi-arch/x86/avx512_vnni.cpp
```diff
@@ -0,0 +1,32 @@
+/*
```
ahuber21: Renamed to avx512_vnni.cpp

include/svs/multi-arch/x86/avx2.cpp
ahuber21: Added

cmake/generate-dispatch-surface.cmake
ahuber21: Added a few more functions.

cmake/dispatch-surface.cmake
```diff
@@ -0,0 +1,119 @@
...
+#
+# On the instruction budgets in particular: `skylake-avx512` rather than
+# `cascadelake` for the AVX512 level, because that level promises AVX-512 F/BW/DQ
+# and nothing more. `cascadelake` also enables AVX512-VNNI, which the compiler is
+# then free to emit into kernels that run on any AVX512F host -- a Skylake-SP
+# among them, where `vpdpbusd` does not exist. VNNI is its own level instead, so
+# that the budget and the promise line up.
+set(SVS_ISA_LEVELS
+    "AVX2|haswell|avx2"
+    "AVX512|skylake-avx512|avx512"
```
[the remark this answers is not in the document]
ahuber21: This was not introduced by this PR, in fact, adding VNNI improves things slightly. Opening an issue...

tests/multi-arch/dispatch-checks/check-dispatch-instructions.cmake
```diff
@@ -0,0 +1,144 @@
...
+    set(${out_var} ${count} PARENT_SCOPE)
+endfunction()
+
+set(errors 0)
+set(summary)
+
+foreach(class IN LISTS svs_required)
+    svs_count_class(count "${class}")
+    if(count EQUAL 0)
+        message("${SVS_OBJECT} contains no ${class} instructions.")
```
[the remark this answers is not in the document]
ahuber21: Fixed

PR #376, opened 2026-08-25
ci(faiss): Gate the LVQ/LeanVec examples on the build, not the CPU vendor

ahuber21:
LVQ/LeanVec examples _should_ fail on non-Intel _or_ if lib was built from public sources only. The second part was missing, which lead to test failures on Intel runner with public sources. This PR fixes it.

Supersedes #314, which keyed off the `-public-only` artifact name and was closed as a sporadic issue.

[The comment](https://github.com/intel/ScalableVectorSearch/pull/314#issuecomment-4202305315) states
>we should be supporting clean fallback on non-Intel hardware and that's what these tests are validating

This is not correct. We do in fact want a failure. So all good.

### DIFF
--- a/.github/scripts/test-faiss.sh
+++ b/.github/scripts/test-faiss.sh
@@ -66,18 +66,19 @@ storage_kind_rejection="The specified storage kind is not compatible with the"
 
 # A nonzero exit alone would let a broken example, a missing library or a segfault pass as expected.
 expect_storage_kind_rejection() {
+  local label="$1"
   local output status
-  output=$("$@" 2>&1) && status=0 || status=$?
+  output=$("${@:2}" 2>&1) && status=0 || status=$?
   echo "$output"
   if [ "$status" -eq 0 ]; then
-    echo "UNEXPECTED: $* succeeded although $lvq_leanvec_missing"
+    echo "UNEXPECTED: $label succeeded although $lvq_leanvec_missing"
     return 1
   fi
   if ! printf '%s\n' "$output" | grep -qF "$storage_kind_rejection"; then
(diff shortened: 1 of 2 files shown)


.github/scripts/test-faiss.sh
```diff
@@ -51,25 +51,46 @@ echo " FAISS C++ tests: "
...
+fi
+
+# The trailing class name differs between bindings/cpp/src/vamana_index.cpp and
+# dynamic_vamana_index.cpp, so matching the full message would pin the examples to one index kind.
+storage_kind_rejection="The specified storage kind is not compatible with the"
+
+# A nonzero exit alone would let a broken example, a missing library or a segfault pass as expected.
+expect_storage_kind_rejection() {
+  local label="$1"
+  shift
```
ethanglaser: Some of the bash syntax is unfamiliar so might be useful to comment. For instance, `set first arg to label and shift to use remaining args below`

ethanglaser (approved):

.github/scripts/test-faiss.sh
ahuber21: @copilot refactor to avoid using `shift`

.github/scripts/test-faiss.sh
```diff
@@ -66,31 +66,27 @@ storage_kind_rejection="The specified storage kind is not compatible with the"
 
 # A nonzero exit alone would let a broken example, a missing library or a segfault pass as expected.
 expect_storage_kind_rejection() {
-  local label="$1"
```
ahuber21: @copilot this will change output. Restore same behavior as the version that had `shift`, but don't use `shift`. Assigning `label="$1"` is fine.

PR #380, opened 2026-09-02
[C API] Implement custom allocator interface and related functions for memory management

### LINKED ISSUE #342: [C API] SVS custom allocator api support
---

## Problem

SVS allocates from its own private C++ heap. PostgreSQL cannot observe, account for, or bound those
allocations. PostgreSQL memory cap has no effect on Vamana index builds because the SVS graph
structure is invisible to PG memory contexts.

A custom allocator hook in the SVS C API would let the PG extension intercept every SVS allocation
with a byte counter, turning the governance model from a pre-flight estimate into real-time
enforcement.

---

## Existing SVS Allocator Infrastructure

SVS's C++ layer already has a well-designed allocator abstraction — this proposal asks to surface it
through the C API.

**`svs::lib::Allocator<T>`** (`include/svs/lib/memory.h`) — default STL-compatible allocator, uses
`::operator new(n, std::align_val_t(alignof(T)))`. This is what `svs::c_runtime::MaybeBlockedAlloc`
defaults to when no custom allocator is provided.

**`svs::HugepageAllocator<T>`** (`include/svs/core/allocator.h`) — allocates via `mmap` with
hugepage support (1GB → 2MB → 4KB fallback). Used as the default allocator for graph adjacency
lists: `SimpleGraph<Idx, HugepageAllocator<Idx>>`.

**`svs::AllocatorInterface` / `AllocatorHandle<T>`** (`include/svs/core/allocator.h`) — a
type-erased runtime-polymorphic allocator base class already present in the C++ layer. Has
`allocate(n)`, `deallocate(ptr, n)`, `clone()`, and `rebind_*()` virtual methods. This is the
natural C++ target for the C API wrapper to call into.

**`svs_threadpool_interface`** (`include/svs/c_api/svs_c.h`, lines 76–90) — the existing custom
thread pool C API. The allocator API should follow exactly the same vtable + self-pointer pattern.

---

## Proposed C API

The allocator API is modelled on `svs_threadpool_interface` — the existing pattern SVS already uses
for caller-supplied implementations.

### `svs_c.h` additions

```c
/* -----------------------------------------------------------------------
 * Custom allocator interface
 *
 * All function pointers MUST be thread-safe — SVS build and search paths
 * call them concurrently from multiple threads.
 *
 * `self` is the opaque caller-owned state pointer passed through unchanged
 * on every call. Use it to carry a byte counter, a lock, or a pool handle.
 *
 * `size` is passed to free/aligned_free so callers do not need to
 * maintain a side-table of allocation sizes (standard free(ptr) does not
 * communicate size back to the caller).
 * ----------------------------------------------------------------------- */

// clang-format off
struct svs_allocator_interface_ops {
    void *(*malloc)       (void *self, size_t size);
    void  (*free)         (void *self, void *ptr,  size_t size);
    void *(*realloc)      (void *self, void *ptr,  size_t old_size, size_t new_size);

    /* aligned_alloc is mandatory — SVS uses SIMD types requiring alignment > 8.
     * HugepageAllocator uses operator new(n, align_val_t), CacheAlignedAllocator
     * uses 64-byte alignment for thread-local storage. Both must be intercepted. */
    void *(*aligned_alloc)(void *self, size_t alignment, size_t size);
    void  (*aligned_free) (void *self, void *ptr, size_t alignment, size_t size);
};
// clang-format on

struct svs_allocator_interface {
    struct svs_allocator_interface_ops ops;
    void *self;
};

typedef struct svs_allocator_interface *svs_allocator_i;
```

### New builder function

```c
/**
 * Attach a custom allocator to a builder.
 *
 * Must be called after svs_index_builder_create() and before svs_index_build()
 * or svs_index_build_dynamic().
 *
 * SVS does not copy the allocator_interface struct — the caller must keep it
 * alive for the lifetime of any index built or loaded with this builder.
 *
 * Passing NULL restores the default allocator (svs::lib::Allocator /
 * HugepageAllocator depending on the data structure).
 *
 * @param builder   Builder handle to configure.
 * @param allocator Caller-supplied allocator interface, or NULL to reset.
 * @param out_err   Error output handle.
 */
SVS_API void svs_index_builder_set_allocator(
    svs_index_builder_h          builder,
    svs_allocator_i              allocator,   /* NULL → reset to default */
    svs_error_h                  out_err
);

/**
 * Attach a custom allocator to a loaded index (for search-path allocations —
 * thread-local candidate buffers, search history pools, etc.).
 *
 * Must be called before the first search on this index handle.
 *
 * @param index     Index handle to configure.
 * @param allocator Caller-supplied allocator interface, or NULL to reset.
 * @param out_err   Error output handle.
 */
SVS_API void svs_index_set_allocator(
    svs_index_h                  index,
    svs_allocator_i              allocator,
    svs_error_h                  out_err
);
```

---

## Required SVS C++ Changes

The C API additions require threading the allocator through the existing C++ dispatch path:

**`IndexBuilder::build()` / `build_dynamic()`** (`bindings/c/src/index_builder.hpp`) must forward
the allocator to `dispatch_vamana_index_build()` and `dispatch_dynamic_vamana_index_build()`.

**`dispatch_vamana_index_build()`** (`bindings/c/src/dispatcher_vamana.hpp`) must accept an
optional `AllocatorHandle` and pass it down to graph construction:

```cpp
svs::Vamana dispatch_vamana_index_build(
    const svs::index::vamana::VamanaBuildParameters& build_params,
    svs::data::ConstSimpleDataView<float> data,
    const Storage* storage,
    svs::DistanceType distance_type,
    svs::threads::ThreadPoolHandle pool,
    std::optional<svs::AllocatorHandle<std::byte>> allocator = std::nullopt  // NEW
);
```

**`SimpleGraph` construction** (`include/svs/core/graph.h`, `default_graph()`) currently defaults
to `HugepageAllocator<Idx>`. With the custom allocator, the C API wrapper builds a
`svs::AllocatorHandle` backed by the caller's vtable and passes it through.

**The C → C++ bridge** in `svs_c.cpp` wraps the C vtable in a thin `AllocatorInterface` subclass:

```cpp
class CAllocatorBridge : public svs::AllocatorInterface {
    svs_allocator_i iface_;
public:
    CAllocatorBridge(svs_allocator_i iface) : iface_(iface) {}
    void* allocate(size_t n) override {
        return iface_->ops.malloc(iface_->self, n);
    }
    void deallocate(void* ptr, size_t n) override {
        iface_->ops.free(iface_->self, ptr, n);
    }
    // clone(), rebind_float(), rebind_float16() forward to same vtable
};
```

---

## Why `size` Must Be Passed to `free`

Standard C `free(ptr)` does not communicate size back to the caller. Without `size` in the free
callback, the counter can only go up — you'd need to maintain a side-table of `ptr → size` mappings
to decrement it on free, which requires a lock and a hash map inside every free call.

Passing `size` to `free` and `aligned_free` makes the counter trivially correct without any
bookkeeping. SVS already knows the allocation size at the deallocation call site (it tracks it to
call sized `operator delete`), so this is no extra burden on SVS.

---

## Why `aligned_alloc` Cannot Be Omitted

SVS uses two distinct aligned allocation paths that would bypass a `malloc`-only hook:

1. **`svs::lib::Allocator<T>`** — uses `::operator new(n, std::align_val_t(alignof(T)))` for any
   type with alignment > `__STDCPP_DEFAULT_NEW_ALIGNMENT__` (16 bytes on x86). SIMD types
   (AVX-512) have 64-byte alignment requirements.

2. **`CacheAlignedAllocator<T, 64>`** — used by `SequentialTLS` for thread-local build buffers.
   Uses `::operator new(bytes, std::align_val_t{64})` directly.

Both must be intercepted to get an accurate byte count.

---

## Open Questions for SVS Team

1. **NULL return handling:** SVS must handle `NULL` returns from the allocator by propagating
   `SVS_ERROR_OUT_OF_MEMORY` rather than dereferencing or `abort()`ing. Is this already guaranteed
   by the C++ exception model, or does it require explicit `if (!ptr) throw std::bad_alloc()`
   guards at each allocation site?

2. **`realloc` usage:** Does SVS ever call `realloc`? If yes, the bridge must implement it. If no,
   we can assert-fail on that vtable slot to catch regressions.

3. **Allocator lifetime for loaded indexes:** `svs_index_set_allocator` is proposed for search-path
   allocations on already-loaded indexes. Are search-path allocations per-query (freed at end of
   search) or retained (thread-local buffers that persist)? This affects whether the counter
   accurately reflects live usage or cumulative usage.

4. **`AllocatorHandle` rebind:** The existing `AllocatorInterface` has `rebind_float()` and
   `rebind_float16()` virtual methods. The C bridge needs to implement these. Should they return a
   new `CAllocatorBridge` backed by the same vtable, or does SVS expect a different allocator for
   different element types?

---

rfsaliev:
This pull request introduces a flexible custom allocator interface to the SVS C API, enabling users to specify memory allocation strategies—including default, huge page, or user-defined custom allocators—when building or loading indexes. The changes propagate allocator support throughout the C API, C++ runtime, and index builder infrastructure, ensuring that memory management can be tailored for performance or integration needs.

Key changes include:

**C API: Allocator Interface and Integration**
- Added a new `svs_allocator_kind` enum and a detailed `svs_allocator_interface` struct, allowing users to select or implement custom allocators. Macros and typedefs were also introduced for ease of use. [[1]](diffhunk://#diff-ae9945586d4cafd685f3a042580fa5316156f4ae00c843f502593658ca47fb65R192-R250) [[2]](diffhunk://#diff-ae9945586d4cafd685f3a042580fa5316156f4ae00c843f502593658ca47fb65R486-R493)
- New API functions: `svs_index_builder_set_allocator` and `svs_index_builder_set_allocator_custom` allow setting the allocator kind or providing a custom allocator implementation for index builders.

**C++ Runtime: Custom Allocator Support**
- Implemented a `CustomAllocator` class template in C++ that wraps the C API allocator interface, validates its structure, and provides allocation/deallocation logic. A factory function (`make_custom_allocator_handle`) is provided for safe construction.
- Updated data builder templates (e.g., for LeanVec, LVQ, Simple, and SQ) to accept allocator handles, ensuring allocator propagation through all builder specializations. [[1]](diffhunk://#diff-811943a906bc2b6d50496a4ccb2d9e6ecda932c1b5aba09246bf0e1e0dd36dfeL143-R144) [[2]](diffhunk://#diff-71f42890738a05fa27b2b395a74c294ed8f00801ed65f35b6e43ca973b53da29L156-R157) [[3]](diffhunk://#diff-f3ba11abdd57c387a19589945930e35c4fd7593343cd74ce1538771f07228d87L94-R97) [[4]](diffhunk://#diff-610950847add2fb9ebb89679db6f18c3f1d1faa946d98a17bc0e49a972e953d7L96-R99)

**Index Building and Loading: Allocator Propagation**
- Modified dynamic and static Vamana index build/load dispatchers and implementations to accept and use allocator handles, ensuring all memory allocations can use the specified allocator. This includes proper rebinding for different value types. [[1]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R48-R70) [[2]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R81-R99) [[3]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R136) [[4]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R155) [[5]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R164) [[6]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R175) [[7]](diffhunk://#diff-f73875f8bbc75a8b601362b974c3ad46ac52582c2b47e5a7ee51538cf2e80933R184) [[8]](diffhunk://#diff-037b3945d8fbb9964a2256d45e1ec3b2e30094386fced7aa1b0a5221a0ebbe37R40) [[9]](diffhunk://#diff-037b3945d8fbb9964a2256d45e1ec3b2e30094386fced7aa1b0a5221a0ebbe37R50) [[10]](diffhunk://#diff-4d0f273362e0b22bbf59bbd72315c7e68b9205b002495c4d063412be95cd1555L44-R56)

**Memory Estimation Improvements**
- Updated the memory estimation logic for dynamic Vamana indexes to use the correct allocator and blocking parameters, improving accuracy and consistency with actual index construction.

These changes lay the groundwork for advanced memory management strategies in SVS, improving performance tuning and integration flexibility for users needing custom allocation behavior.

### DIFF
--- a/bindings/c/include/svs/c/svs_c.h
+++ b/bindings/c/include/svs/c/svs_c.h
@@ -825,9 +825,10 @@ SVS_API bool svs_index_builder_set_allocator(
 /// @param allocator The custom allocator interface
 /// @param out_err An optional error handle to capture errors
 /// @return true on success, false on failure
-/// @remarks The builder copies @p allocator (the interface struct and its ops table) by
-/// value before returning, so the caller may free or modify @p allocator and its ops table
-/// once this call returns.
+/// @remarks The builder copies @p allocator and its ops table by value, so those two
+/// objects may be freed or modified after this call returns. The object referenced by
+/// @p allocator->self is not copied and must outlive the builder and every index built or
+/// loaded with it.
 SVS_API bool svs_index_builder_set_allocator_custom(
     svs_index_builder_h builder, svs_allocator_i allocator, svs_error_h out_err /*=NULL*/
 );
--- a/bindings/c/include/svs/c/svs_c.h
+++ b/bindings/c/include/svs/c/svs_c.h
@@ -189,6 +189,65 @@ struct svs_threadpool_interface {
     void* self;
 };
 
+/// @brief Allocators used for memory management in the SVS C API.
+enum svs_allocator_kind {
+    SVS_ALLOCATOR_KIND_DEFAULT = 0,
+    SVS_ALLOCATOR_KIND_HUGE_PAGE = 1,
+    SVS_ALLOCATOR_KIND_CUSTOM = 2
+};
+
+/// @brief Operations table for a custom allocator interface
+/// @remarks The user must ensure that the allocator implementation is thread-safe and
+/// that the provided function pointers remain valid for the lifetime of the allocator
+/// interface.
+/// @var svs_allocator_interface_ops::version
+///   Version of the allocator interface.
+/// @var svs_allocator_interface_ops::struct_size
+///   Size of the structure, used for versioning and compatibility checks.
+/// @var svs_allocator_interface_ops::allocate
+///   Function pointer to allocate memory with the specified size and alignment.
+///   @param self Pointer to the allocator instance.
+///   @param size Size of the memory to allocate in bytes.
+///   @param alignment Alignment requirement for the allocated memory in bytes.
+///   @param out_err Handle to capture any error that occurs during allocation. User code
+///   may call svs_error_set() to set the error code and message if an error occurs.
+///   @return Pointer to the allocated memory, or NULL if allocation fails.
+/// @var svs_allocator_interface_ops::deallocate
+///   Function pointer to deallocate memory previously allocated by the allocator.
+///   @param self Pointer to the allocator instance.
+///   @param ptr Pointer to the memory to deallocate.
+///   @param size Size of the memory to deallocate in bytes.
+///   @param alignment Alignment requirement for the allocated memory in bytes.
+struct svs_allocator_interface_ops {
+    uint32_t version;
+    size_t struct_size;
+    void* (*allocate)(void* self, size_t size, size_t alignment, svs_error_h out_err);
+    void (*deallocate)(void* self, void* ptr, size_t size, size_t alignment);
+};
+
+/// @brief Macro to create a user-defined allocator interface operations structure
+/// @param allocate_func Function pointer to allocate memory with the specified size and
+/// alignment
+/// @param deallocate_func Function pointer to deallocate memory previously allocated by the
+/// allocator
+#define SVS_INIT_ALLOCATOR_OPS(allocate_func, deallocate_func)     \
+    {                                                              \
+        .version = SVS_C_API_VERSION,                              \
+        .struct_size = sizeof(struct svs_allocator_interface_ops), \
+        .allocate = &allocate_func, .deallocate = &deallocate_func \
+    }
+
+/// @brief Structure representing a custom allocator interface
+/// @var svs_allocator_interface::ops
+///   Function pointers for the allocator operations.
+/// @var svs_allocator_interface::self
+///   Pointer to the user-defined allocator instance. This pointer is passed to the
+///   function pointers in @p ops when they are called.
+struct svs_allocator_interface {
+    struct svs_allocator_interface_ops* ops;
+    void* self;
+};
+
 /// @brief Operations table for a custom ID filter interface
 /// @remarks The user must ensure that the ID filter implementation is thread-safe and
 /// that the provided function pointers remain valid for the lifetime of the ID filter
@@ -425,10 +484,14 @@ typedef enum svs_algorithm_type svs_algorithm_type_t;
 typedef enum svs_data_type svs_data_type_t;
 typedef enum svs_storage_kind svs_storage_kind_t;
 typedef enum svs_threadpool_kind svs_threadpool_kind_t;
+typedef enum svs_allocator_kind svs_allocator_kind_t;
 
 typedef struct svs_threadpool_interface_ops svs_threadpool_ops_t;
 typedef struct svs_threadpool_interface svs_threadpool_t;
 typedef struct svs_threadpool_interface* svs_threadpool_i;
+typedef struct svs_allocator_interface_ops svs_allocator_ops_t;
+typedef struct svs_allocator_interface svs_allocator_t;
+typedef struct svs_allocator_interface* svs_allocator_i;
 
 typedef struct svs_id_filter_interface_ops svs_id_filter_ops_t;
 typedef struct svs_id_filter_interface svs_id_filter_t;
@@ -746,6 +809,27 @@ SVS_API bool svs_index_builder_set_threadpool(
 /// call returns.
 SVS_API bool svs_index_builder_set_threadpool_custom(
     svs_index_builder_h builder, svs_threadpool_i pool, svs_error_h out_err /*=NULL*/
+);
+
+/// @brief Set the allocator configuration for the index builder
+/// @param builder The index builder handle
+/// @param kind The kind of allocator to use
+/// @param out_err An optional error handle to capture errors
+/// @return true on success, false on failure
+SVS_API bool svs_index_builder_set_allocator(
+    svs_index_builder_h builder, svs_allocator_kind_t kind, svs_error_h out_err /*=NULL*/
+);
+
+/// @brief Set the custom allocator for the index builder
+/// @param builder The index builder handle
+/// @param allocator The custom allocator interface
+/// @param out_err An optional error handle to capture errors
+/// @return true on success, false on failure
+/// @remarks The builder copies @p allocator (the interface struct and its ops table) by
+/// value before returning, so the caller may free or modify @p allocator and its ops table
+/// once this call returns.
+SVS_API bool svs_index_builder_set_allocator_custom(
+    svs_index_builder_h builder, svs_allocator_i allocator, svs_error_h out_err /*=NULL*/
 );
 
 /// @brief Estimate the memory usage of an index based on the builder configuration and
--- a/bindings/c/include/svs/c/svs_c.h
+++ b/bindings/c/include/svs/c/svs_c.h
@@ -192,8 +192,9 @@ struct svs_threadpool_interface {
 /// @brief Allocators used for memory management in the SVS C API.
 enum svs_allocator_kind {
     SVS_ALLOCATOR_KIND_DEFAULT = 0,
-    SVS_ALLOCATOR_KIND_HUGE_PAGE = 1,
-    SVS_ALLOCATOR_KIND_CUSTOM = 2
+    SVS_ALLOCATOR_KIND_SIMPLE = 1,
+    SVS_ALLOCATOR_KIND_HUGE_PAGE = 2,
+    SVS_ALLOCATOR_KIND_CUSTOM = 3
 };
 
 /// @brief Operations table for a custom allocator interface
--- a/bindings/c/src/allocator.hpp
+++ b/bindings/c/src/allocator.hpp
@@ -15,8 +15,15 @@
  */
 #pragma once
 
+#include "svs/c/svs_c.h"
+
+#include "error.hpp"
+
+#include <svs/core/allocator.h>
 #include <svs/core/data/simple.h>
+#include <svs/lib/float16.h>
 #include <svs/lib/memory.h>
+#include <svs/lib/meta.h>
 
 namespace svs {
 namespace c_runtime {
@@ -25,5 +32,69 @@ template <typename T, bool UseBlocked, typename Allocator = svs::lib::Allocator<
 using MaybeBlockedAlloc =
     std::conditional_t<UseBlocked, svs::data::Blocked<Allocator>, Allocator>;
 
+template <typename T> class CustomAllocator : public svs::AllocatorInterface {
+  public:
+    using value_type = T;
+
+    static void validate(const svs_allocator_i allocator) {
+        if (allocator == nullptr) {
+            throw std::invalid_argument("Custom allocator pointer cannot be null.");
+        }
+        if (allocator->ops == nullptr) {
+            throw std::invalid_argument("Custom allocator interface is not initialized.");
+        }
+        if (allocator->ops->allocate == nullptr || allocator->ops->deallocate == nullptr) {
+            throw std::invalid_argument(
+                "Custom allocator interface has null function pointers."
+            );
+        }
+    }
+
+    CustomAllocator(const svs_allocator_ops_t& ops, void* self)
+        : ops_(ops)
+        , self_(self) {}
+
+    void* allocate(size_t n) override {
+        svs_error_desc err{SVS_ERROR_UNKNOWN, "Unknown error in custom allocator allocate"};
+
+        auto result = ops_.allocate(self_, n, alignof(T), &err);
+        if (result == nullptr) {
+            throw std::runtime_error(
+                "Custom allocator failed to allocate memory: (" + std::to_string(err.code) +
+                ") " + err.message
+            );
+        }
+        return result;
+    }
+
+    void deallocate(void* p, size_t n) override {
+        ops_.deallocate(self_, p, n, alignof(T));
+    }
+
+    AllocatorInterface* clone() const override {
+        return new CustomAllocator<T>(ops_, self_);
+    }
+
+    AllocatorInterface* rebind_to(DataType type) const override {
+        return svs::lib::match(
+            AllocatorInterface::rebind_types{},
+            type,
+            [this]<typename Tag>(svs::lib::Type<Tag>) -> AllocatorInterface* {
+                return new CustomAllocator<Tag>(ops_, self_);
+            }
+        );
+    }
+
+  private:
+    svs_allocator_ops_t ops_;
+    void* self_;
+};
+
+template <typename T = std::byte>
+AllocatorHandle<T> make_custom_allocator_handle(const svs_allocator_i allocator) {
+    CustomAllocator<T>::validate(allocator);
+    return AllocatorHandle<T>{
+        std::make_unique<CustomAllocator<T>>(*allocator->ops, allocator->self)};
+}
 } // namespace c_runtime
 } // namespace svs
--- a/bindings/c/src/allocator.hpp
+++ b/bindings/c/src/allocator.hpp
@@ -57,7 +57,7 @@ template <typename T> class CustomAllocator : public svs::AllocatorInterface {
     void* allocate(size_t n) override {
         svs_error_desc err{SVS_ERROR_UNKNOWN, "Unknown error in custom allocator allocate"};
 
-        auto result = ops_.allocate(self_, n, alignof(T), &err);
+        auto result = ops_.allocate(self_, n * sizeof(T), alignof(T), &err);
         if (result == nullptr) {
             throw std::runtime_error(
                 "Custom allocator failed to allocate memory: (" + std::to_string(err.code) +
@@ -68,7 +68,7 @@ template <typename T> class CustomAllocator : public svs::AllocatorInterface {
     }
 
     void deallocate(void* p, size_t n) override {
-        ops_.deallocate(self_, p, n, alignof(T));
+        ops_.deallocate(self_, p, n * sizeof(T), alignof(T));
     }
 
     AllocatorInterface* clone() const override {
--- a/bindings/c/src/allocator.hpp
+++ b/bindings/c/src/allocator.hpp
@@ -43,6 +43,16 @@ template <typename T> class CustomAllocator : public svs::AllocatorInterface {
         if (allocator->ops == nullptr) {
             throw std::invalid_argument("Custom allocator interface is not initialized.");
         }
+        if (allocator->ops->version > svs_get_version()) {
+            throw std::invalid_argument(
+                "Custom allocator interface version is not supported."
+            );
+        }
+        if (allocator->ops->struct_size < sizeof(svs_allocator_ops_t)) {
+            throw std::invalid_argument(
+                "Incompatible custom allocator interface struct size."
+            );
+        }
         if (allocator->ops->allocate == nullptr || allocator->ops->deallocate == nullptr) {
             throw std::invalid_argument(
                 "Custom allocator interface has null function pointers."
@@ -59,7 +69,7 @@ template <typename T> class CustomAllocator : public svs::AllocatorInterface {
 
         auto result = ops_.allocate(self_, n * sizeof(T), alignof(T), &err);
         if (result == nullptr) {
-            throw std::runtime_error(
+            throw svs::c_runtime::out_of_memory(
                 "Custom allocator failed to allocate memory: (" + std::to_string(err.code) +
                 ") " + err.message
             );
--- a/bindings/c/src/allocator.hpp
+++ b/bindings/c/src/allocator.hpp
@@ -32,79 +32,123 @@ template <typename T, bool UseBlocked, typename Allocator = svs::lib::Allocator<
 using MaybeBlockedAlloc =
     std::conditional_t<UseBlocked, svs::data::Blocked<Allocator>, Allocator>;
 
-template <typename T> class CustomAllocator : public svs::AllocatorInterface {
-  public:
-    using value_type = T;
+class AllocatorBuilder {
+    template <typename T> class CustomAllocator {
+      public:
+        using value_type = T;
 
-    static void validate(const svs_allocator_i allocator) {
-        if (allocator == nullptr) {
-            throw std::invalid_argument("Custom allocator pointer cannot be null.");
-        }
-        if (allocator->ops == nullptr) {
-            throw std::invalid_argument("Custom allocator interface is not initialized.");
-        }
-        if (allocator->ops->version > svs_get_version()) {
-            throw std::invalid_argument(
-                "Custom allocator interface version is not supported."
-            );
-        }
-        if (allocator->ops->struct_size < sizeof(svs_allocator_ops_t)) {
-            throw std::invalid_argument(
-                "Incompatible custom allocator interface struct size."
-            );
-        }
-        if (allocator->ops->allocate == nullptr || allocator->ops->deallocate == nullptr) {
-            throw std::invalid_argument(
-                "Custom allocator interface has null function pointers."
-            );
-        }
-    }
-
-    CustomAllocator(const svs_allocator_ops_t& ops, void* self)
-        : ops_(ops)
-        , self_(self) {}
-
-    void* allocate(size_t n) override {
-        svs_error_desc err{SVS_ERROR_UNKNOWN, "Unknown error in custom allocator allocate"};
-
-        auto result = ops_.allocate(self_, n * sizeof(T), alignof(T), &err);
-        if (result == nullptr) {
-            throw svs::c_runtime::out_of_memory(
-                "Custom allocator failed to allocate memory: (" + std::to_string(err.code) +
-                ") " + err.message
-            );
-        }
-        return result;
-    }
-
-    void deallocate(void* p, size_t n) override {
-        ops_.deallocate(self_, p, n * sizeof(T), alignof(T));
-    }
-
-    AllocatorInterface* clone() const override {
-        return new CustomAllocator<T>(ops_, self_);
-    }
-
-    AllocatorInterface* rebind_to(DataType type) const override {
-        return svs::lib::match(
-            AllocatorInterface::rebind_types{},
-            type,
-            [this]<typename Tag>(svs::lib::Type<Tag>) -> AllocatorInterface* {
-                return new CustomAllocator<Tag>(ops_, self_);
+        static void validate(const svs_allocator_i allocator) {
+            if (allocator == nullptr) {
+                throw std::invalid_argument("Custom allocator pointer cannot be null.");
             }
-        );
+            if (allocator->ops == nullptr) {
+                throw std::invalid_argument("Custom allocator interface is not initialized."
+                );
+            }
+            if (allocator->ops->version > svs_get_version()) {
+                throw std::invalid_argument(
+                    "Custom allocator interface version is not supported."
+                );
+            }
+            if (allocator->ops->struct_size < sizeof(svs_allocator_ops_t)) {
+                throw std::invalid_argument(
+                    "Incompatible custom allocator interface struct size."
+                );
+            }
+            if (allocator->ops->allocate == nullptr ||
+                allocator->ops->deallocate == nullptr) {
+                throw std::invalid_argument(
+                    "Custom allocator interface has null function pointers."
+                );
+            }
+        }
+
+        CustomAllocator(const svs_allocator_ops_t& ops, void* self)
+            : ops_(ops)
+            , self_(self) {}
+
+        template <typename U> friend class CustomAllocator;
+
+        template <typename U>
+        CustomAllocator(const CustomAllocator<U>& other)
+            : ops_(other.ops_)
+            , self_(other.self_) {}
+
+        template <typename U> CustomAllocator& operator=(const CustomAllocator<U>& other) {
+            ops_ = other.ops_;
+            self_ = other.self_;
+            return *this;
+        }
+
+        value_type* allocate(size_t n) {
+            svs_error_desc err{
+                SVS_ERROR_UNKNOWN, "Unknown error in custom allocator allocate"};
+
+            auto result = ops_.allocate(self_, n * sizeof(T), alignof(T), &err);
+            if (result == nullptr) {
+                throw svs::c_runtime::out_of_memory(
+                    "Custom allocator failed to allocate memory: (" +
+                    std::to_string(err.code) + ") " + err.message
+                );
+            }
+            return static_cast<value_type*>(result);
+        }
+
+        void deallocate(value_type* p, size_t n) {
+            ops_.deallocate(self_, p, n * sizeof(T), alignof(T));
+        }
+
+      private:
+        svs_allocator_ops_t ops_;
+        void* self_;
+    };
+
+    svs_allocator_kind kind_;
+    // Owned copy of the user's allocator vtable; `self_` is referenced only.
+    svs_allocator_ops_t user_ops_;
+    void* user_self_;
+
+  public:
+    AllocatorBuilder(svs_allocator_kind kind = SVS_ALLOCATOR_KIND_DEFAULT)
+        : kind_(kind) {
+        switch (kind_) {
+            case SVS_ALLOCATOR_KIND_DEFAULT:
+            case SVS_ALLOCATOR_KIND_SIMPLE:
+            case SVS_ALLOCATOR_KIND_HUGE_PAGE:
+                break;
+            case SVS_ALLOCATOR_KIND_CUSTOM:
+                throw std::invalid_argument(
+                    "SVS_ALLOCATOR_KIND_CUSTOM requires a custom allocator interface."
+                );
+            default:
+                throw std::invalid_argument("Unknown allocator kind.");
+        }
     }
 
-  private:
-    svs_allocator_ops_t ops_;
-    void* self_;
-};
+    AllocatorBuilder(svs_allocator_i allocator)
+        : kind_(SVS_ALLOCATOR_KIND_CUSTOM)
+        , user_ops_(*allocator->ops)
+        , user_self_(allocator->self) {
+        CustomAllocator<std::byte>::validate(allocator);
+    }
 
-template <typename T = std::byte>
-AllocatorHandle<T> make_custom_allocator_handle(const svs_allocator_i allocator) {
-    CustomAllocator<T>::validate(allocator);
-    return AllocatorHandle<T>{
-        std::make_unique<CustomAllocator<T>>(*allocator->ops, allocator->self)};
-}
+    svs_allocator_kind kind() const { return kind_; }
+
+    template <typename T = std::byte> svs::AllocatorHandle<T> build() const {
+        // For now, default allocator is simple allocator - can be changed in the future.
+        switch (kind_) {
+            case SVS_ALLOCATOR_KIND_DEFAULT:
+            case SVS_ALLOCATOR_KIND_SIMPLE:
+                return svs::make_allocator_handle(svs::lib::Allocator<T>{});
+            case SVS_ALLOCATOR_KIND_HUGE_PAGE:
+                return svs::make_allocator_handle(svs::HugepageAllocator<T>{});
+            case SVS_ALLOCATOR_KIND_CUSTOM:
+                return svs::make_allocator_handle(CustomAllocator<T>{user_ops_, user_self_}
+                );
+            default:
+                throw std::invalid_argument("Unknown allocator kind.");
+        }
+    }
+};
 } // namespace c_runtime
 } // namespace svs
--- a/bindings/c/src/allocator.hpp
+++ b/bindings/c/src/allocator.hpp
@@ -150,6 +150,22 @@ class AllocatorBuilder {
                 throw std::invalid_argument("Unknown allocator kind.");
         }
     }
+
+    template <typename T = uint32_t> svs::AllocatorHandle<T> build_for_graph() const {
+        // Default graph allocator is hugepage allocator - can be changed in the future.
+        switch (kind_) {
+            case SVS_ALLOCATOR_KIND_SIMPLE:
+                return svs::make_allocator_handle(svs::lib::Allocator<T>{});
+            case SVS_ALLOCATOR_KIND_DEFAULT:
+            case SVS_ALLOCATOR_KIND_HUGE_PAGE:
+                return svs::make_allocator_handle(svs::HugepageAllocator<T>{});
+            case SVS_ALLOCATOR_KIND_CUSTOM:
+                return svs::make_allocator_handle(CustomAllocator<T>{user_ops_, user_self_}
+                );
+            default:
+                throw std::invalid_argument("Unknown allocator kind.");
+        }
+    }
 };
 } // namespace c_runtime
 } // namespace svs
--- a/bindings/c/src/allocator.hpp
+++ b/bindings/c/src/allocator.hpp
@@ -105,8 +105,8 @@ class AllocatorBuilder {
 
     svs_allocator_kind kind_;
     // Owned copy of the user's allocator vtable; `self_` is referenced only.
-    svs_allocator_ops_t user_ops_;
-    void* user_self_;
+    svs_allocator_ops_t user_ops_{};
+    void* user_self_ = nullptr;
 
   public:
     AllocatorBuilder(svs_allocator_kind kind = SVS_ALLOCATOR_KIND_DEFAULT)
@@ -126,10 +126,11 @@ class AllocatorBuilder {
     }
 
     AllocatorBuilder(svs_allocator_i allocator)
-        : kind_(SVS_ALLOCATOR_KIND_CUSTOM)
-        , user_ops_(*allocator->ops)
-        , user_self_(allocator->self) {
+        : kind_(SVS_ALLOCATOR_KIND_CUSTOM) {
         CustomAllocator<std::byte>::validate(allocator);
+        // Copy the 'vtable' since the user may free the original.
+        user_ops_ = *allocator->ops;
+        user_self_ = allocator->self;
     }
 
     svs_allocator_kind kind() const { return kind_; }
--- a/bindings/c/src/dispatcher_dynamic_vamana.cpp
+++ b/bindings/c/src/dispatcher_dynamic_vamana.cpp
@@ -45,6 +45,7 @@ svs::DynamicVamana build_dynamic_vamana_index(
     DataBuilder builder,
     Distance D,
     svs::threads::ThreadPoolHandle pool,
+    const AllocatorHandle<std::byte>& allocator_handle,
     size_t blocksize_bytes
 ) {
     svs::data::BlockingParameters block_params;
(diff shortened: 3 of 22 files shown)


bindings/c/src/index_builder.hpp
```diff
@@ -57,7 +58,8 @@ struct IndexBuilder {
         , dimension(dimension)
         , algorithm(std::move(algorithm))
         , storage(std::make_shared<StorageSimple>(SVS_DATA_TYPE_FLOAT32))
-        , pool_builder{} {}
+        , pool_builder{}
+        , allocator_handle{make_allocator_handle(svs::lib::Allocator<std::byte>{})} {}
```
yuejiaointel: This handle also becomes the **graph's** allocator now that [`dispatcher_dynamic_vamana.cpp#L62-L63`](https://github.com/intel/ScalableVectorSearch/blob/0c1c149a581719a4171139ffc34c286a5a5a0e7b/bindings/c/src/dispatcher_dynamic_vamana.cpp#L62-L63) routes the graph through it. Before this PR the graph was `SimpleBlockedGraph<uint32_t>`, hardcoded to `HugepageAllocator` ([`graph.h#L451`](https://github.com/intel/ScalableVectorSearch/blob/0c1c149a581719a4171139ffc34c286a5a5a0e7b/include/svs/core/graph/graph.h#L451)), so C API users silently lose huge pages on the graph — the structure traversed on every search.

`HugepageAllocator` already falls back to 4 KB pages when huge pages are unavailable ([`allocator.h#L95-L99`](https://github.com/intel/ScalableVectorSearch/blob/0c1c149a581719a4171139ffc34c286a5a5a0e7b/include/svs/core/allocator.h#L95)), so defaulting to it keeps the previous behaviour:

```suggestion
        , allocator_handle{make_allocator_handle(svs::HugepageAllocator<std::byte>{})} {}
```

This also switches the *data* default to huge pages. If you'd rather keep data on `lib::Allocator`, the graph needs its own handle instead.

include/svs/core/allocator.h
```diff
@@ -621,27 +641,16 @@ template <typename T> class AllocatorHandle {
...
-    template <typename U>
+    template <HasDataType U>
     AllocatorHandle& operator=(const AllocatorHandle<U>& other)
-        requires std::is_same_v<T, Float16> && (!std::is_same_v<U, T>)
+        requires(!std::is_same_v<T, U>) && (lib::in<U>(AllocatorInterface::rebind_types{}))
     {
-        impl_.reset(other.impl_->rebind_float16());
+        impl_.reset(other.impl_->rebind_to(datatype_v<T>));
         return *this;
     }
```
ethanglaser: **The rebind constraint is on the wrong template parameter.** The `requires` clause validates the *source* type `U`, but the dispatch uses the *target*: `rebind_to(datatype_v<T>)`. It should be `HasDataType T` and `lib::in<T>(AllocatorInterface::rebind_types{})`.

As written, rebinding *to* a type that isn't in `rebind_types` satisfies the constraint, compiles, and then throws `ANNEXCEPTION("Type {} is not supported for this operation!")` out of `lib::match` at runtime (`include/svs/lib/meta.h:143`) — whereas the old `rebind_float` / `rebind_float16` overloads made that a hard compile error. Same issue on the `operator=` constraint at line 651. This looks like an unintended side effect of the generalization rather than a deliberate trade.

include/svs/core/allocator.h
```diff
@@ -600,6 +617,9 @@ template <typename T> class AllocatorHandle {
   public:
     using value_type = T;
 
+    explicit AllocatorHandle(std::unique_ptr<AllocatorInterface> impl)
```
ethanglaser: **This constructor is unguarded.** The existing templated constructor just below enforces `std::is_same_v<typename Impl::value_type, T>`; this one enforces nothing. So:

```cpp
AllocatorHandle<float>{std::make_unique<CustomAllocator<std::byte>>(ops, self)};
```

compiles, and the resulting handle hands out `float*` for buffers sized in bytes — a 4x heap overflow. `make_custom_allocator_handle` uses it correctly, but this is a public constructor on a public header, so nothing stops the next caller from getting it wrong. Could it be private with `make_custom_allocator_handle` as a friend, or could the value type be carried on `AllocatorInterface` so this can actually be checked?

include/svs/core/allocator.h
```diff
@@ -556,21 +557,34 @@ concept Allocator = HasValueType<Alloc> && std::is_copy_constructible_v<Alloc>
...
+
     virtual ~AllocatorInterface() = default;
     virtual void* allocate(size_t n) = 0;
     virtual void deallocate(void* ptr, size_t n) = 0;
 
     // covariant return type
     virtual AllocatorInterface* clone() const = 0;
-    virtual AllocatorInterface* rebind_float() const = 0;
-    virtual AllocatorInterface* rebind_float16() const = 0;
+    virtual AllocatorInterface* rebind_to(DataType type) const = 0;
```
ethanglaser: Worth an explicit call: `AllocatorInterface` sits under `@defgroup core_allocators_public`, and this changes its vtable — two pure virtuals removed, one added. Any out-of-tree subclass breaks at compile time, and an old shared library paired with a new header is UB. Probably acceptable given who consumes this today, but I'd rather it be a deliberate decision than an implicit one.

bindings/c/src/index_builder.hpp
ethanglaser: Two concerns on this one line.

**Hugepages are silently dropped for the graph.** Before this PR every C API graph allocation used hugepages: static build via `Vamana::build`'s default `Allocator = HugepageAllocator<uint32_t>` (`include/svs/orchestrators/vamana.h:588`), static load via `GraphLoader{}` -> `SimpleGraph<uint32_t, HugepageAllocator<uint32_t>>`, and dynamic build via the deduction guide to `SimpleBlockedGraph<uint32_t>`, which is hardcoded to `BlockedData<Idx, Dynamic, HugepageAllocator<Idx>>` (`include/svs/core/graph/graph.h:450`). All three now take this handle, which defaults to `svs::lib::Allocator<std::byte>` — plain `::operator new`. Vector data was already `lib::Allocator` via `MaybeBlockedAlloc`, so only the graph regresses, but that's the TLB-sensitive part of the search hot path, and it now applies to every default-configured index rather than being opt-in. `HugepageAllocator` already falls back to 4 KiB pages when hugepages aren't available (`force_ = false`), so pointing `SVS_ALLOCATOR_KIND_DEFAULT` at it should be safe. Is the change intentional? If so, could the PR description carry a benchmark number?

**Storing a materialized handle loses the kind.** `ThreadPoolBuilder`, in this same directory, stores `{kind, num_threads, user_ops_, user_self_}` and constructs the pool in `build()` (`bindings/c/src/threadpool.hpp:150`). An `AllocatorBuilder` following that pattern would keep the kind available to everything else that needs it — see my comment on `dispatcher_dynamic_vamana.cpp` for the concrete consequence — and would leave room for a `svs_index_builder_get_allocator` later.

bindings/c/src/dispatcher_dynamic_vamana.cpp
```diff
@@ -175,24 +195,26 @@ svs::index::vamana::MemoryBreakdown dispatch_dynamic_vamana_memory_estimate(
     size_t blocksize_bytes
 ) {
     svs::index::vamana::MemoryBreakdown breakdown{};
+
+    // Graph size
     // Graph: SimpleBlockedData<uint32_t> with num_vectors rows and (max_degree + 1)
     // cols; the +1 slot stores the per-node neighbor count.
     using index_type = uint32_t;
+    using graph_allocator_type = svs::data::Blocked<svs::lib::Allocator<index_type>>;
```
ethanglaser, line 203: This is the fallout from `IndexBuilder` storing a live handle rather than a builder: the estimator has no way to know which allocator was configured, so it has to hardcode one. The result is that `svs_index_builder_estimate_memory_dynamic` is wrong for `SVS_ALLOCATOR_KIND_HUGE_PAGE` (2 MiB rounding is unaccounted for) and wrong for any custom allocator. `dispatch_vamana_memory_estimate` (~line 159 of `dispatcher_vamana.cpp`, unchanged here) is allocator-blind entirely — it still uses `SimpleDataBuilder<index_type>{}`.

Note that if the motivation for defaulting to `lib::Allocator` was to make the estimate line up with `within_1pct(tracker.live_bytes, memory_usage)` in the new tests, that's fixing the estimator's inaccuracy by degrading the runtime. Threading the *kind* through instead fixes both.

bindings/c/src/svs_c.cpp
```diff
@@ -716,6 +719,54 @@ SVS_API bool svs_index_builder_estimate_search_memory_dynamic(
...
+                    );
+                    break;
+                }
+                case SVS_ALLOCATOR_KIND_HUGE_PAGE: {
+                    builder->impl->set_allocator_handle(
+                        svs::make_allocator_handle(svs::HugepageAllocator<std::byte>{})
+                    );
+                    break;
+                }
+                default:
```
ethanglaser: `SVS_ALLOCATOR_KIND_CUSTOM` is declared in the public enum but unreachable through this function — it lands in `default:` and reports "Invalid allocator kind", which is actively misleading since it's a documented enumerator. Either give it a targeted message pointing at `svs_index_builder_set_allocator_custom`, or drop the enumerator from the header.

Related: calling `svs_index_builder_set_allocator(builder, DEFAULT, ...)` after `svs_index_builder_set_allocator_custom` silently discards the custom allocator, with no way for the caller to observe which one is currently active.

include/svs/index/vamana/dynamic_index.h
```diff
@@ -201,6 +201,56 @@ class MutableVamanaIndex {
         translator_.insert(external_ids, threads::UnitRange<Idx>(0, external_ids.size()));
     }
 
+    template <typename ExternalIds, typename ThreadPoolProto>
```
ethanglaser, line 204: This constructor plus `auto_dynamic_build` largely duplicate the existing build-from-scratch constructor at line 257 — same parameter verification, same member assignment, same two `VamanaBuilder::construct` calls. Can the older one delegate to this one so there's a single dynamic build path?

Two concrete reasons beyond tidiness: the two paths have already diverged on prefetch parameters (see my comment on line 245), and the older one carries a redundant full graph reallocation that delegating would delete outright (see my comment on line 230 — ~1.3 GB allocated and thrown away for a 10M-node degree-32 graph).

include/svs/index/vamana/dynamic_index.h
```diff
@@ -201,6 +201,56 @@ class MutableVamanaIndex {
...
+        prune_to_ = build_parameters_.prune_to;
+        use_full_search_history_ = build_parameters_.use_full_search_history;
+        // Perform graph construction.
+        auto builder = VamanaBuilder(
+            graph_,
+            data_,
+            distance_,
+            build_parameters_,
+            threadpool_,
+            extensions::estimate_prefetch_parameters(data_),
```
ethanglaser, line 249: This is where the two dynamic build paths diverge: the pre-existing build-from-scratch constructor passes `GreedySearchPrefetchParameters{sp.prefetch_lookahead_, sp.prefetch_step_}`, while this one uses `extensions::estimate_prefetch_parameters(data_)`. This one matches `VamanaIndex` (`include/svs/index/vamana/index.h:435`), so it's the more consistent of the two — but two dynamic build entry points with different prefetch behavior is exactly the drift the duplication invites.

include/svs/index/vamana/dynamic_index.h
```diff
@@ -201,6 +201,56 @@ class MutableVamanaIndex {
...
+              external_ids,
+              std::move(threadpool_proto),
+              std::move(logger)} {
+        if (graph_.n_nodes() != data_.size()) {
+            throw ANNEXCEPTION("Wrong sizes!");
+        }
+        build_parameters_ = parameters;
+        // Verify and set defaults before using the parameters to set other member
+        // variables.
+        verify_and_set_default_index_parameters(build_parameters_, distance_function);
```
ethanglaser: `distance_function` was moved into the delegated constructor on line 220, so this reads a moved-from object and then copies it (the callee takes `Dist` by value, `include/svs/index/vamana/index.h:1084`).

It's benign today — `verify_and_set_default_index_parameters` only does `std::decay_t<decltype(distance_function)>` and branches on the type, never reading any state — but that's an implementation detail of the callee, not something the signature guarantees. The delegated constructor has already stored the functor in `distance_` (line 195), so:

```diff
-        verify_and_set_default_index_parameters(build_parameters_, distance_function);
+        verify_and_set_default_index_parameters(build_parameters_, distance_);
```

Same type, same by-value cost, correct by construction. The identical pattern exists at `index.h:428` and at `dynamic_index.h:277`, if you want to fix it at the source.

---

Separately, while checking the older constructor's comment on line 280 ("Set graph again as verify function might change graph_max_degree parameter") — that comment is not just stale, it's justifying a redundant full graph allocation:

```diff
-        // Set graph again as verify function might change graph_max_degree parameter
-        graph_ = Graph{data_.size(), build_parameters_.graph_max_degree};
```

The init list already builds `graph_(Graph{data.size(), parameters.graph_max_degree})`, `build_parameters_` is copy-initialized from the same `parameters`, and `verify_and_set_default_index_parameters` only writes `max_candidate_pool_size`, `prune_to`, and `alpha` — it never touches `graph_max_degree`. So line 281 constructs a provably identical graph and move-assigns it over the original. The temporary is fully constructed before the assignment, so both graphs are live at once: for 10M nodes at degree 32 that's ~1.3 GB allocated, peaking at ~2.6 GB, then freed — and on the default hugepage path those are `MAP_POPULATE` pages, so they get faulted in before being discarded.

Pre-existing rather than introduced here, and it disappears entirely if the older constructor delegates to the new one (see my comment on line 204).

include/svs/index/vamana/dynamic_index.h
```diff
@@ -201,6 +201,56 @@ class MutableVamanaIndex {
...
+
+        assert(build_parameters_.graph_max_degree == graph_.max_degree());
```
ethanglaser: This is a debug-only check on a caller-supplied precondition, two lines below a real `throw ANNEXCEPTION("Wrong sizes!")` for the same class of error. A mismatched `graph_max_degree` in a release build will silently produce a corrupt index. I'd make it a throw for consistency with line 225.

include/svs/core/graph.h
```diff
@@ -31,9 +31,10 @@ namespace svs {
 ///
 /// @tparam Idx The type used to encode nodes in the graph.
 ///
-template <typename Idx = uint32_t> struct GraphLoader {
+template <typename Idx = uint32_t, typename Allocator = HugepageAllocator<Idx>>
```
ethanglaser, line 36: Now that `GraphLoader` is allocator-parameterized, it's a near-duplicate of `svs::python::UnspecializedGraphLoader` (`bindings/python/include/svs/python/core.h:98`) — same "path + allocator, rebind to `uint32_t`, `SimpleGraph<uint32_t, other>::load`" shape. The same overlap now exists between `graphs::SimpleBlockedGraph<Idx>` and `SimpleGraph<Idx, Blocked<A>>`.

I don't see an obvious shared home for these given how differently the two binding layers are structured, so this is probably a note for later rather than something to act on in this PR — but flagging it so it doesn't get forgotten.

bindings/c/include/svs/c/svs_c.h
```diff
@@ -748,6 +811,28 @@ SVS_API bool svs_index_builder_set_threadpool_custom(
...
+SVS_API bool svs_index_builder_set_allocator(
+    svs_index_builder_h builder, svs_allocator_kind_t kind, svs_error_h out_err /*=NULL*/
+);
+
+/// @brief Set the custom allocator for the index builder
+/// @param builder The index builder handle
+/// @param allocator The custom allocator interface
+/// @param out_err An optional error handle to capture errors
+/// @return true on success, false on failure
+/// @remarks The builder copies @p allocator and its ops table by value, so those two
```
ethanglaser, line 829: This `@remarks` block is good on lifetime, but silent on *scope*, which is the first thing a caller of this API needs to know. The custom allocator covers vector data and the graph. It does not cover:

- **ID translation metadata.** `IDTranslator` holds two `tsl::robin_map`s with default allocators and no allocator parameter (`include/svs/core/translation.h:440-441`). At robin_map's 0.5 default load factor that's roughly 64 bytes per point across the two maps — for a 10M-point dynamic index, ~600 MB against a degree-32 graph of ~1.3 GB. Not a rounding error, and the new test encodes the gap as an expectation (`within_1pct(tracker.live_bytes, data_bytes + graph_bytes)`, deliberately excluding `metadata_bytes`).
- **Search scratchspace.** `SearchBuffer` uses `threads::CacheAlignedAllocator` (`include/svs/index/vamana/search_buffer.h:113`). Small and per-thread, but it means the guarantee doesn't hold even for pure search workloads.
- **LeanVec training data.** `svs_leanvec_training_data_build` creates an independent handle before the builder is configured (`bindings/c/src/svs_c.cpp:422`), so it structurally can't route through this allocator.
- **The memory estimators** (see my note on `dispatcher_dynamic_vamana.cpp`).

Callers reaching for a custom allocator are typically doing arena allocation, NUMA pinning, or hard memory accounting, and for all three "most of the bytes, unspecified which" is hard to build against. Could the boundary be stated here explicitly, with the metadata gap called out as the material one?

bindings/c/src/allocator.hpp
```diff
@@ -25,5 +32,79 @@ template <typename T, bool UseBlocked, typename Allocator = svs::lib::Allocator<
...
+        }
+        if (allocator->ops == nullptr) {
+            throw std::invalid_argument("Custom allocator interface is not initialized.");
+        }
+        if (allocator->ops->version > svs_get_version()) {
+            throw std::invalid_argument(
+                "Custom allocator interface version is not supported."
+            );
+        }
+        if (allocator->ops->struct_size < sizeof(svs_allocator_ops_t)) {
```
ethanglaser: This rejects `struct_size` *smaller* than the current `svs_allocator_ops_t`, which defeats the purpose of shipping a `struct_size` field — an older, smaller caller struct is precisely the case that field exists to support. `svs_threadpool_interface_ops` and `svs_id_filter_interface_ops` don't check it at all. Three interfaces with three different conventions is worse than any one of them; worth settling on a single rule and documenting it next to the field declaration in `svs_c.h`.

bindings/c/src/allocator.hpp
```diff
@@ -25,5 +32,79 @@ template <typename T, bool UseBlocked, typename Allocator = svs::lib::Allocator<
...
+            );
+        }
+    }
+
+    CustomAllocator(const svs_allocator_ops_t& ops, void* self)
+        : ops_(ops)
+        , self_(self) {}
+
+    void* allocate(size_t n) override {
+        svs_error_desc err{SVS_ERROR_UNKNOWN, "Unknown error in custom allocator allocate"};
```
ethanglaser: This constructs an `svs_error_desc` on *every* allocation, and its message is 41 characters — past SSO, so it heap-allocates each time on a path that runs once per block. Worse, under genuine OOM the error-reporting path itself allocates. Constructing the descriptor lazily, only on the `nullptr` return, would be both cheaper and more robust.

bindings/c/src/error.hpp
```diff
@@ -97,6 +97,11 @@ class unsupported_hw : public std::runtime_error {
     using std::runtime_error::runtime_error;
 };
 
+class out_of_memory : public std::runtime_error {
```
ethanglaser, line 100: Non-blocking, but worth a follow-up at least.

`wrap_exceptions` has no `catch (const std::bad_alloc&)`, and `std::bad_alloc` derives from `std::exception` rather than `std::runtime_error` — so it bypasses the `std::runtime_error` clause below and lands in the catch-all as `SVS_ERROR_UNKNOWN`. After this PR that means the same condition reports three different codes depending on the configured allocator:

| Allocator | Failure path | Reported code |
| --- | --- | --- |
| `SVS_ALLOCATOR_KIND_CUSTOM` | `ops_.allocate` returns `nullptr` -> `out_of_memory` | `SVS_ERROR_OUT_OF_MEMORY` |
| `SVS_ALLOCATOR_KIND_DEFAULT` | `lib::Allocator` -> `::operator new` throws `std::bad_alloc` | `SVS_ERROR_UNKNOWN` |
| `SVS_ALLOCATOR_KIND_HUGE_PAGE` | `mmap` fails -> `ANNEXCEPTION` (`include/svs/core/allocator.h:147`) | `SVS_ERROR_GENERIC` |

Note that `SVS_ERROR_OUT_OF_MEMORY = 3` has been in the public enum (`svs_c.h:53`) since before this PR with nothing ever producing it — this makes it reachable for the first time, but only on the path a default-configured caller won't take. Anyone branching on `SVS_ERROR_OUT_OF_MEMORY` to implement a retry-with-smaller-index policy still won't see it from the default allocator.

Adding `catch (const std::bad_alloc& ex)` -> `SVS_ERROR_OUT_OF_MEMORY`, placed above the `std::exception` clause, would make the code mean what its name says. Deriving `out_of_memory` from `std::bad_alloc` instead also works, but it's more awkward than it sounds: `std::bad_alloc` has no string constructor and its `what()` is fixed, so you'd have to store the message and override `what()` yourself. Keeping `std::runtime_error` as the base and just adding the missing catch clause is the smaller change.

bindings/c/tests/c_api_dynamic_index.cpp
```diff
@@ -738,6 +744,122 @@ CATCH_TEST_CASE("C API Dynamic Index Memory", "[c_api][index][memory][dynamic]")
...
+            bool success = svs_index_builder_set_threadpool(
+                local_builder, SVS_THREADPOOL_KIND_NATIVE, NUM_THREADS, error
+            );
+            CATCH_REQUIRE(success);
+
+            bool ok = svs_index_builder_set_allocator(local_builder, kind, error);
+            CATCH_REQUIRE(ok);
+            CATCH_REQUIRE(svs_error_ok(error));
+
+            // Use big blocksize to avoid many HugePage allocations.
```
ethanglaser, line 765: The comment says "use big blocksize to avoid many HugePage allocations," but `hugepage_mmap` sets `MAP_POPULATE` on Linux (`include/svs/core/allocator.h:136`), so `1 << 30` means a ~1 GiB data block *and* a ~1 GiB graph block get eagerly faulted in — roughly 2 GiB resident, to build an index over 256 vectors of dimension 32. That's a lot to put on CI runners for a smoke test. A modest blocksize gets the same coverage.

bindings/c/tests/c_api_dynamic_index.cpp
```diff
@@ -388,14 +389,19 @@ CATCH_TEST_CASE("C API Dynamic Index Memory", "[c_api][index][memory][dynamic]")
 
     svs_error_h error = svs_error_create();
```
ethanglaser, line 391: Changing from `100, 100` to `32, 50`, plus the 4-thread pool added at line 400, changes the configuration for *every pre-existing section* in this test case — including the memory-accounting ones, which now measure something different than they did before. Why is the change necessary?

bindings/c/tests/c_api_index.cpp
```diff
@@ -1181,6 +1181,154 @@ CATCH_TEST_CASE("C API Index Memory Management", "[c_api][index][memory]") {
...
+        success = svs_index_get_memory_usage(index, &memory_usage, error);
+        CATCH_REQUIRE(success);
+        CATCH_REQUIRE(svs_error_ok(error));
+
+        // For the static Vamana index both the data and the graph are allocated through
+        // the builder's allocator, so the bytes still held by the custom allocator must
+        // match the graph + data portion of the breakdown (the tiny metadata entry point
+        // is not routed through the allocator). The same must hold against the pre-build
+        // estimate.
+        auto within_1pct = [](size_t a, size_t b) {
```
ethanglaser, line 1314: `svs_index_get_memory_usage` returns `breakdown.total()` = `graph_bytes + data_bytes + metadata_bytes` (`include/svs/index/vamana/index.h:188`), but the tracker only ever sees data and graph — so this tolerance is silently absorbing `metadata_bytes`, which is exactly what the comment above says is excluded. Comparing against `svs_index_get_memory_breakdown`'s `data_bytes + graph_bytes` instead, as the dynamic test does at `bindings/c/tests/c_api_dynamic_index.cpp:854`, would let this be an exact equality and would document its own intent.

That said, the dynamic test *already* excludes metadata and still uses `within_1pct` (line 840) — with fixed `DIMENSION` / `GRAPH_DEGREE` / `BLOCK_SIZE` the block counts there are deterministic, so what residual is that one absorbing?

bindings/c/tests/c_api_test_utils.h
```diff
@@ -199,3 +200,54 @@ check_training_data_support(svs_leanvec_training_data_h training_data, svs_error
...
+inline void* tracking_allocator_allocate(
+    void* self, size_t size, size_t alignment, svs_error_h out_err
+) {
+    auto* tracker = static_cast<TrackingAllocator*>(self);
+    // std::aligned_alloc requires a power-of-two alignment that is at least
+    // alignof(std::max_align_t) and a size that is a multiple of that alignment.
+    size_t align =
+        alignment < alignof(std::max_align_t) ? alignof(std::max_align_t) : alignment;
+    size_t rounded = ((size + align - 1) / align) * align;
+    void* ptr = std::aligned_alloc(align, rounded);
```
ethanglaser: `std::aligned_alloc` isn't provided by MSVC's STL (`_aligned_malloc` / `_aligned_free` are the equivalents there), and this header does get compiled on Windows — `svs_c_config.h` has `_WIN32` handling. Test-only, so low priority, but it will break that build.

bindings/c/include/svs/c/svs_c.h
rfsaliev: Unfortunately, implementing "customizable allocator" calability to listed coponents requires significant refactoring for the whole SVS project - which is out of this PR scope. This PR contains C API exports for existing SVS functionality + minimum required changes in SVS core code.

bindings/c/src/allocator.hpp
rfsaliev: The `struct_size` field is reserved for future potential changes in the interface.
For now, at this particular moment, it should be at least the size of the current structure.
In future, in case of extended `svs_allocator_ops_t` this condition to be modified.

bindings/c/src/dispatcher_dynamic_vamana.cpp
rfsaliev: It seems like both memory breakdown accounting and memory estimation to be fixed to properly support HugepageAllocator case.
Making the new "issue" record because it is pretty complex and out-of the scope of the current PR.

rfsaliev: Reported as new issue #386

bindings/c/tests/c_api_dynamic_index.cpp
rfsaliev: These changes just to reduce test execution time.

bindings/c/src/error.hpp
rfsaliev: Added `std::bad_alloc` handling to `wrap_exception`.
Did not touch `HugepageAllcocator` to keep this PR as small as possible - just the minimum required set of changes.

bindings/c/src/index_builder.hpp
rfsaliev: 1. Using hugepage allocations for the graph causes significant memory "wasting" in case if dynamic index is built with small blocksize - `HugepageAllocator` always allocate at least 1 huge page for every block.
  * There is no benchmarks yet for C API.
  * Other APIs kept unchanged - HugepageAllocator is defaulted for a graph.

2. Implemented `AllocatorBuilder` - thank you for the note.

bindings/c/src/svs_c.cpp
rfsaliev: Refactored using `AllocatorBuilder`
Please review again

bindings/c/tests/c_api_dynamic_index.cpp
rfsaliev: Done.

bindings/c/tests/c_api_index.cpp
rfsaliev: Sorry, do not understand the question.
_For now_, `VamanaIndex::get_memory_breakdown()` accounts only `entry_point_` in `metadata_bytes`.
In case if memory breakdown computation will be changed significantly, e.g. new big enough metadata will be added to `VamanaIndex`
, and memory allocation for such metadata will not use the provided allocator - we will be able to catch this by the test.
Such issue is already detected in `/c/tests/c_api_dynamic_index.cpp`, but marked as __TODO__ comment because implementing custom allocators for `MutableVamanaIndex`'s metadata is the effort and is out of the scope for this PR - which exposes existing SVS functionality in C API.

include/svs/core/allocator.h
rfsaliev: Removed.

include/svs/core/allocator.h
rfsaliev: Fixed.

include/svs/index/vamana/dynamic_index.h
rfsaliev: Changes in this PR are additive as much as possible to avoid potential issues in existing code.
This why I've decided to add a new .ctor() rater than modifying the existing one.
Agree that `MutableVamanaIndex` to be refactored as far as many other components, but modifying existing functionality is out of the scope for this PR.
Instead, I would propose implement such changes in a separate PR.

include/svs/index/vamana/dynamic_index.h
rfsaliev: To get `sp`, the pre-existing .ctor() calls `get_search_parameters()` which returns `search_parameters_` which initialized by `construct_default_search_parameters()` which calls `estimate_prefetch_parameters(data_)`.
So, technically, prefetch parameters are equal in both cases.

include/svs/index/vamana/dynamic_index.h
rfsaliev: 1. Thank you for the good point - fixed.
2. Let's implement such fixes in a separate PR.

include/svs/index/vamana/dynamic_index.h
rfsaliev: Agree, fixed.

bindings/c/tests/c_api_dynamic_index.cpp
ethanglaser: How about using `BLOCK_SIZE` instead of `1 << 12` and dropping the comment? The other build calls in this test case pass `BLOCK_SIZE`, so the literal is the only magic blocksize, so once it's `BLOCK_SIZE` there's nothing left to explain.

bindings/c/src/allocator.hpp
```diff
@@ -25,5 +32,124 @@ template <typename T, bool UseBlocked, typename Allocator = svs::lib::Allocator<
...
+        user_ops_ = *allocator->ops;
+        user_self_ = allocator->self;
+    }
+
+    svs_allocator_kind kind() const { return kind_; }
+
+    template <typename T = std::byte> svs::AllocatorHandle<T> build() const {
+        // For now, default allocator is simple allocator - can be changed in the future.
+        switch (kind_) {
+            case SVS_ALLOCATOR_KIND_DEFAULT:
```
ethanglaser, line 141: `SVS_ALLOCATOR_KIND_DEFAULT` and `SVS_ALLOCATOR_KIND_SIMPLE` resolve to the same allocator, so `DEFAULT` has no distinct meaning today. More importantly, since the dispatchers now take the *graph* allocator from this handle, the default path changes graph storage from `HugepageAllocator<uint32_t>` to `lib::Allocator<uint32_t>` — previously `Vamana::build` and `GraphLoader` supplied the hugepage default, and `auto_dynamic_build` still defaults to `Blocked<HugepageAllocator<uint32_t>>`.

That's a silent search-latency change for any caller that never calls `svs_index_builder_set_allocator`. Is that intended? Mapping `DEFAULT` to `HugepageAllocator` would keep the C API default aligned with the C++ default and give the enum value a purpose.

bindings/c/src/dispatcher_dynamic_vamana.cpp
ethanglaser, line 203: This hard-codes `lib::Allocator<index_type>` for the graph (and `estimate_data_size_blocked` below doesn't take an allocator either), while `build_dynamic_vamana_index` above now derives both from the caller's `AllocatorHandle`.

With `SVS_ALLOCATOR_KIND_HUGE_PAGE` — which rounds every allocation up to 2 MiB — or a custom allocator with its own rounding, `svs_index_builder_estimate_memory` and `svs_index_get_memory_breakdown` will disagree for the same builder config, which defeats the point of pairing them.

Threading the `AllocatorHandle` into `dispatch_dynamic_vamana_memory_estimate` the same way as the build/load paths seems like the natural fix; failing that, worth documenting that estimates assume the simple allocator.

include/svs/core/allocator.h
```diff
@@ -556,21 +557,34 @@ concept Allocator = HasValueType<Alloc> && std::is_copy_constructible_v<Alloc>
 
 class AllocatorInterface {
   public:
+    // The types that allocator handle can rebind to.
+    // Order here is for performance - most common types are listed first.
+    using rebind_types = lib::Types<
```
ethanglaser, line 562: Nice simplification over the two hand-written rebinds.

One consequence worth confirming: `rebind_to` instantiates `lib::rebind_allocator_t<T, Impl>` for all 13 types in this list for every `AllocatorImpl<Impl>`, up from 2. Any allocator that cannot rebind to one of them (e.g. one with a constrained `value_type`) becomes a hard compile error at `AllocatorImpl` instantiation rather than at the point of use. Probably fine, but worth a sanity check against `HugepageAllocator`, the MMap allocators, and the new `CustomAllocator` — and an eye on compile time.

include/svs/index/vamana/dynamic_index.h
```diff
@@ -201,6 +201,60 @@ class MutableVamanaIndex {
...
+        Data data,
+        Idx entry_point,
+        Dist distance_function,
+        const ExternalIds& external_ids,
+        ThreadPoolProto threadpool_proto,
+        // Optional logger parameter
+        svs::logging::logger_ptr logger = svs::logging::get()
+    )
+        : MutableVamanaIndex{
+              std::move(graph),
```
ethanglaser, line 217: Minor and pre-existing: the constructor this delegates to reads its moved-from parameter at line 198 — `construction_window_size_{2 * graph.max_degree()}` is evaluated after `graph_{std::move(graph)}`.

Benign today because `DenseArray` retains its dimensions after a move, but this new constructor puts it on the dynamic *build* path, so switching it to `graph_.max_degree()` would be worth doing while you're here.

ethanglaser (commented):
Thanks for quick resolutions, a couple additional observations and follow-ups but otherwise looking good. Is the plan to merge this into #364 or merge #364 to main first?

include/svs/core/allocator.h
rfsaliev: There are C API Memory Management tests which imply rebinding between `byte`, `float`, `float16` and `uint32_t`.

bindings/c/src/dispatcher_dynamic_vamana.cpp
rfsaliev: Agree, but to resolve this issue, the actual memory breakdown calculation in `MutableVamanaIndex::get_memory_breakdown()` have to be rewritten to make it allocator-aware.
Such fix is seemed like is out of this PR scope.

bindings/c/src/allocator.hpp
rfsaliev: Mapping DEFAILT to `HugepageAllocator` would lead silent search-latency and memory consumption as well, because vectors dataset uses `svs::lib::Allocator` by default.
Generally, there is allocators mismatch between datasets and graph in the current SVS code: datasets use `lib::Allocator` by default when graph always use `HugepageAllocator`.
Using `HugepageAllocator` by default introduce the risk of "wasting" significant amount of memory for small indices or small blocksizes, because it may (depending on OS config) reserve `1GB` page for __every__ allocation.

bindings/c/tests/c_api_dynamic_index.cpp
rfsaliev: In case of using `BLOCK_SIZE`, the index will be built with 4 blocks of vectors + 4 blocks of graph. In case of `HugepageAllocator`, depending on OS configuration, it may take 8GB memory - which is pretty big for CI runners.
My typo: here have to be `1 << 21`.

include/svs/index/vamana/dynamic_index.h
rfsaliev: Thank you.
Made "while I'm here' fixes at lines 198 and 282

ethanglaser (approved):
LGTM!

bindings/c/src/allocator.hpp
```diff
@@ -25,5 +32,124 @@ template <typename T, bool UseBlocked, typename Allocator = svs::lib::Allocator<
...
+            case SVS_ALLOCATOR_KIND_SIMPLE:
+                return svs::make_allocator_handle(svs::lib::Allocator<T>{});
```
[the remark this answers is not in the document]
rfsaliev: Refactored the allocator handling for `SVS_ALLOCATOR_KIND_DEFAULT` to set `lib::Allocator` for data and `HugepageAllocator` for graph.

bindings/c/tests/c_api_index.cpp
```diff
@@ -1181,6 +1181,154 @@ CATCH_TEST_CASE("C API Index Memory Management", "[c_api][index][memory]") {
...
+            svs_index_builder_set_allocator_custom(builder, nullptr, error) == false
+        );
+        CATCH_REQUIRE(svs_error_get_code(error) == SVS_ERROR_INVALID_ARGUMENT);
+
+        success = svs_index_builder_set_allocator_custom(builder, &allocator, error);
+        CATCH_REQUIRE(success);
+        CATCH_REQUIRE(svs_error_ok(error));
+
+        // Build the index; both data and graph allocations flow through `tracker`.
+        svs_index_h index = svs_index_build(builder, data.data(), NUM_VECTORS, error);
```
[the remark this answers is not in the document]
rfsaliev: Added "Custom Allocator Failure" test section

bindings/c/src/allocator.hpp
rfsaliev: Finally, I found the way to solve the concern:
I've refactored the allocator handling for `SVS_ALLOCATOR_KIND_DEFAULT` to set `lib::Allocator` for data and `HugepageAllocator` for graph.

PR #383, opened 2026-09-07
feat(vamana): Add replace_external_id, a data-preserving external-ID rename

nonirosenfeldredis:
## Summary

`MutableVamanaIndex::delete_entries` + `add_points` was the only existing way to change a
vector's external label -- and that path re-reads and re-inserts the vector's data, and
re-links its graph adjacency, even when the caller only wants a new label for data that
hasn't changed. That's exactly the situation a downstream index-update path faces after a
document update turns out not to have touched the vector: nothing about the stored vector or
its graph position needs to change, only the external ID it's known by.

- `IDTranslator::remap_external_id(from, to)`: the missing symmetric counterpart to the
  existing `remap_internal_id` (used internally by `compact()`) -- a pure O(1) hash-map key
  swap on the external ID, checked by default the same way `insert`/`delete_*` are.
- `MutableVamanaIndex::replace_external_id(old_id, new_id)`: exposes this at the index level,
  alongside the existing `translate_external_id`/`has_id`/`get_datum` methods it's meant to be
  used with. Touches neither the dataset nor the graph.
- `MultiMutableVamanaIndex::replace_external_id(old_label, new_label)`: the multi-value
  counterpart. Pure bookkeeping on `label_to_external_`/
  `external_to_label_` that moves every external id already grouped under `old_label` to
  `new_label` -- the plural case, since a multi-value label can hold more than one id where
  the single-value version only ever has one to move. Same "must exist / must not collide"
  exception semantics as the single-value version.

## Testing

- New "Rename External" section (plus two error-path sections) in
  `tests/svs/core/translation.cpp`.
- New "MutableVamana Index Relabel" test case in `tests/svs/index/vamana/dynamic_index.cpp`,
  verifying the internal id, stored data, and every other id are unaffected by the rename.
- New "MultiMutableVamana Index Replace External ID" test case in
  `tests/svs/index/vamana/multi.cpp`, built on a label with two externals grouped under it so
  the rename has more than one id to move, plus the same two error-path sections.
- Verified all error-path sections actually exercise their checks (not just happen to pass)
  by temporarily disabling each, confirming the throw-assertions fail as expected, then
  restoring it. Same for the multi-value rename itself: verified the new test fails when the
  `external_to_label_` update is skipped, before restoring the fix.

### DIFF
--- a/include/svs/index/vamana/dynamic_index.h
+++ b/include/svs/index/vamana/dynamic_index.h
@@ -463,6 +463,24 @@ class MutableVamanaIndex {
     }
 
     ///
+    /// @brief Rename external ID `old_id` to `new_id`, keeping the same stored vector
+    /// data and adjacency list.
+    ///
+    /// @param old_id The existing external ID to rename.
+    /// @param new_id The external ID to assign. Must not already exist.
+    ///
+    /// Requires that `old_id` exists in the index and `new_id` does not; throws
+    /// otherwise, leaving the index unmodified. Unlike `delete_entries` followed by
+    /// `add_points`, this is a pure bookkeeping operation on the ID translation table --
+    /// it does not touch the underlying dataset or graph.
+    ///
+    /// @see has_id, delete_entries
+    ///
+    void relabelVector(size_t old_id, size_t new_id) {
+        translator_.remap_external_id(old_id, new_id);
+    }
+
+    ///
     /// @brief Get the raw data for external id `e`.
     ///
     auto get_datum(size_t e) const { return data_.get_datum(translate_external_id(e)); }
--- a/include/svs/index/vamana/dynamic_index.h
+++ b/include/svs/index/vamana/dynamic_index.h
@@ -476,7 +476,7 @@ class MutableVamanaIndex {
     ///
     /// @see has_id, delete_entries
     ///
-    void relabelVector(size_t old_id, size_t new_id) {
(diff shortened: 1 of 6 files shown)


include/svs/index/vamana/dynamic_index.h
```diff
@@ -462,6 +462,24 @@ class MutableVamanaIndex {
...
+    /// @param new_id The external ID to assign. Must not already exist.
+    ///
+    /// Requires that `old_id` exists in the index and `new_id` does not; throws
+    /// otherwise, leaving the index unmodified. Unlike `delete_entries` followed by
+    /// `add_points`, this is a pure bookkeeping operation on the ID translation table --
+    /// it does not touch the underlying dataset or graph.
+    ///
+    /// @see has_id, delete_entries
+    ///
+    void relabelVector(size_t old_id, size_t new_id) {
```
rfsaliev: The name `relabelVector` does not match existing naming convention.
Please use different name aka: `replace_external_id`

rfsaliev (changes requested):
Thank you for the contribution.
There are some changes needed:
1. Follow the existing naming convention.
2. Add similar API to the `MutliMuatbleVamanaIndex` in `include/svs/index/vamana/multi.h`

rfsaliev (dismissed):
LGFM

rfsaliev (approved):

PR #390, opened 2026-09-17
fix: forward + default-enable search_buffer_visited_set_ in concurrent MutableVamanaIndex

yuejiaointel:
## What

Two changes, both against PR #369's new `svs::index::vamana::concurrent::MutableVamanaIndex`:

1. **Fix**: `include/svs/concurrent/dynamic_index.h`'s batch-search path constructed `search_buffer_type` with only 2 args (buffer_config_, comparator), silently dropping the 3rd arg (`enable_visited`) — so `search_parameters.search_buffer_visited_set_` never took effect for batch search. Mirrors the already-merged fix for the non-concurrent index (commit 8052bd9, PR #364).
2. **Default flip**: `include/svs/index/vamana/search_params.h`'s `search_buffer_visited_set_` default changed from `false` to `true`. This struct (`VamanaSearchParameters`) is shared/unforked by PR #369 (`concurrent::MutableVamanaIndex::search_parameters_type` is a direct alias to it), so this one edit affects both the concurrent and non-concurrent index identically.

## Why

The visited-set feature has been validated in a large regression campaign (2 datasets x 2 modes x 3 metrics x 4 variants = 48 configs, k=100, recall 0.90-0.95): a consistent 1.2x-1.4x QPS gain, zero recall cost, zero regressions. Flipping the default means callers get the benefit without needing to know the flag exists.

## Note on duplication with #389

`#389` ("Change search_buffer_visited_set_ default from false to true") already proposes the identical default-flip change against `main`, opened separately today. The diff text here is copied verbatim from #389 for consistency — if #389 merges to `main` before this PR's base branch does, rebasing this branch will make that commit a no-op (identical resulting text), not a conflict.

## Verification

- Confirmed via source read that the same missing-3rd-arg bug pattern exists in the concurrent copy (not assumed from the PR title's "98% identical" claim).
- `tests/svs/concurrent/dynamic_index.cpp`, `dynamic_index_2.cpp`, `tests/svs/index/vamana/dynamic_index.cpp`, and `search_parameters.cpp` all recompile cleanly with g++-11 after both changes (confirmed via fresh object-file timestamps).
- Full test-suite link/run not performed — flagging this scope limitation explicitly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mihaic (approved):

yuejiaointel:
Closed #389 as a duplicate — it made the identical search_buffer_visited_set_ default flip directly against main. That change now lives here instead (part of this PR's diff, against dev/eglaser-finegrain-dmitry-copy).

PR #391, opened 2026-09-22
Fix CMAKE_BUILD_TYPE: RelWithDebugInfo -> RelWithDebInfo

Vika-F:
The CMAKE_BUILD_TYPE was fixed in accordance with cmake docs:
https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html#cmake-build-type

Incorrect naming of the build type resulted in the wrong compiler flags, for example, no `-g` flag passed to the build.


ethanglaser (approved):
Nice, thanks for catching this. Looks like a similar update will be needed in innersource.

Vika-F:
@ethanglaser Thank you, but I do not have the right permissions to create a PR in the innersource :-/
Also, I cannot merge this PR.
And I do not think that it is allowed to post links to innersource here [but I might be wrong].

mihaic:
I gave you permission for innersource, @Vika-F. It should become effective today.

